PR #21, opened 2024-02-23
`Pointer::push_back`: Handle empty keys

wngr:
Hey!

When using `Pointer::push_back` with empty strings as keys the `count` was incremented, but the actual `inner` pointer value was not updated.

This PR fixes that.

I'm not sure why the check `!self.inner.ends_with('/')` was present in the first place.

### DIFF
--- a/src/pointer.rs
+++ b/src/pointer.rs
@@ -141,9 +141,7 @@ impl Pointer {
     /// Pushes a `Token` onto the back of this `Pointer`.
     pub fn push_back(&mut self, token: Token) {
         self.count += 1;
-        if !self.inner.ends_with('/') {
-            self.inner.push('/');
-        }
+        self.inner.push('/');
         self.inner.push_str(token.encoded());
     }
 
@@ -167,7 +165,6 @@ impl Pointer {
                     self.inner = String::from("/") + &front;
                     Token::from_encoded(back)
                 }
-                
             })
     }
     /// Removes and returns the first `Token` in the `Pointer` if it exists.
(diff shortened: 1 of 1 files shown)


asmello:
Adding some context.

Per [RFC 6901§5](https://www.rfc-editor.org/rfc/rfc6901#section-5):

> ```
>    For example, given the JSON document
> 
>    {
>       "foo": ["bar", "baz"],
>       "": 0,
>       "a/b": 1,
>       "c%d": 2,
>       "e^f": 3,
>       "g|h": 4,
>       "i\\j": 5,
>       "k\"l": 6,
>       " ": 7,
>       "m~n": 8
>    }
> 
>    The following JSON strings evaluate to the accompanying values:
> 
>     ""           // the whole document
>     "/foo"       ["bar", "baz"]
>     "/foo/0"     "bar"
>     "/"          0
>     "/a~1b"      1
>     "/c%d"       2
>     "/e^f"       3
>     "/g|h"       4
>     "/i\\j"      5
>     "/k\"l"      6
>     "/ "         7
>     "/m~0n"      8
> ```

Note how `/` is a valid JSON pointer that evaluates to a root field keyed by an empty string (`""`), and not the whole document (which is represented as `""`, confusingly enough). Unlike with UNIX-style paths, every forward slash character in JSON pointers is significant and cannot be ignored when constructing paths with `Pointer::push_back`.

chanced:
Yea, the `"/"` being a valid, non-root path confused me when I first wrote it.

The check for `"/"` must have been an artifact I forgot to remove when I realized `""` was actually the root.

Sorry about that and thank you for the fix!

chanced:
This should be available under v0.4.5. Thanks y'all.

PR #25, opened 2024-03-12
`Pointer::pop_back`: Correctly handle empty strings

wngr:
Hi again 😏 !

Empty strings weren't correctly handled in `pop_back`: When calling `pop_back` for the pointer `//`, it would leave an empty string for `Pointer::inner` and `Pointer::count` at 1. Which is obviously wrong and leaves `Pointer` inconsistent.

I also synchronized `fn back` with this, although in this case it was correct.

Next time I create a PR I should probably just contribute a `quickcheck` for this crate directly ;-)

### DIFF
--- a/src/pointer.rs
+++ b/src/pointer.rs
@@ -153,19 +153,12 @@ impl Pointer {
         if self.count > 0 {
             self.count -= 1;
         }
-        self.inner[1..]
-            .rsplit_once('/')
-            .map_or(Some((&self.inner[1..], "")), Option::Some)
-            .map(|(f, b)| (f.to_owned(), b.to_owned()))
-            .map(|(front, back)| {
-                if back.is_empty() {
-                    self.inner = String::default();
-                    Token::from_encoded(front)
-                } else {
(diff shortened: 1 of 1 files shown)


chanced:
Ah, I'm sorry, I missed this somehow.

chanced:
I've published v0.4.7 which includes this and #26  (cc @asmello).

I'm really sorry ya'll have been running into so many issues with this crate. It's quite embarrassing. It was my first project in rust and I haven't shown it the due diligence it deserves.

I plan on circling back to it and cleaning it up once I finish with my JSON schema crate.

Thank you both for all the fixes.

asmello:
That's alright, I think it's really just one class of bugs relating to empty strings and their interaction with slashes. RFC 6901 superficially looks nice and intuitive, but some details are quite surprising and idiosyncratic. The fact that `"/"` points to a field keyed by `""` and `""` points to the root of the document is very confusing and unusual. It's quite easy to make incorrect assumptions.

PR #33, opened 2024-06-12
update CHANGELOG.md

asmello:
Adds the missing notes from #30


asmello:
Oops accidentally included the first commit in #32, but this is the real one @chanced

PR #37, opened 2024-06-14
Rework `Token` type

asmello:
This includes surprisingly few API changes. The main breaks come from the fact `Token` holds a (potentially non-static) lifetime now. I also remove several trait implementations and some methods, in favor of making the distinction between the encoded and decoded representations more explicit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-This is a breaking release including [#30](https://github.com/chanced/jsonptr/pull/30) by [@asmello](https://github.com/asmello)
+This is a breaking release including [#30](https://github.com/chanced/jsonptr/pull/30) and [#37](https://github.com/chanced/jsonptr/pull/37) by [@asmello](https://github.com/asmello)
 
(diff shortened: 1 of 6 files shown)


asmello:
@chanced no urgency, but I think this is ready for a look when you have time.

chanced:
@asmello, dude, you're awesome. This is so much cleaner than anything I would have produced.

I just realized I forgot to squash your last PR. Ah well.

asmello:
> I just realized I forgot to squash your last PR. Ah well.

There's always `git rebase -i` 😂

asmello:
Ah, before I forget, one note since you mentioned you considered `OnceCell` — I assume that was so we could memoize the `decoded` value if it ever gets read. I don't think it's a bad idea, but I also think the `Token` type is a bit of an ephemeral type; people will typically produce it as part of iteration and either use it immediately or convert the decoded value to a `String` for keeping it around long-term.

chanced:
It was. That logic is solid. I'm happy with out it.

chanced:
The idea was to have something like`
```rust
pub struct Token(Internal);

enum Internal<'p> {
    Uncoded(Cow<&str>),
    Encoded { decoded: OnceCell<String>, encoded: Cow<&str> }
}
```

PR #39, opened 2024-06-15
introduce Index type for less error-prone index resolution

### LINKED ISSUE #35: `Token::as_index` OOB behavior
More of a discussion than a issue, perhaps, but I find it unexpected that this is true:

```rust
assert_eq!(Token::new("-").as_index(1), Ok(1));
assert_eq!(Token::new("1").as_index(1), Ok(1));
assert_eq!(Token::new("2").as_index(1), Err(IndexError::OutOfBounds));
```

I can see the utility of the first case, but I'd expect that out-of-bounds numerical indices would always produce an error. It's quite error prone that they don't (it even contradicts the docs, as they're currently written).

If we go by the RFC literally, it defines `-` as an error condition in [section 7](https://datatracker.ietf.org/doc/html/rfc6901#section-7), but doesn't define how to handle the error. In practice we know from sibling standards like [RFC 6902](https://datatracker.ietf.org/doc/html/rfc6902) that using `-` is only an actual error condition in some contexts, so that puts it into a bit of a grey area.

If we were to prioritize semantic correctness, we should probably produce an error for `-` (but a separate error from what would result from a numerical out-of-bounds index). Another option would be to have the `Ok` variant carry an enum value indicating whether the index is internal or refers to the next element:

```rust
enum Index {
    Internal(usize),
    Next(usize)
}
fn as_index(...) -> Result<Index, IndexError>
```

Regardless, I think numerical indices that exceed the given array length should always be considered an error, or never.

asmello:
Fixes #35 

As discussed in #35, this PR introduces a new type `Index` which encodes a parsed index from a `Token` type (or more generally, an RFC 6901-compliant string). 

This type is primarily useful to resolve the `-` token into a concrete numerical index given a known array length. It's also useful for reliably checking bounds against the provided length.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -5,7 +5,6 @@ use alloc::{
     vec::Vec,
 };
 use core::{
-    // error::Error as StdError,
     fmt::{Debug, Display, Formatter},
     num::ParseIntError,
 };
@@ -23,8 +22,6 @@ pub enum Error {
     Unresolvable(UnresolvableError),
     /// Indicates that a Pointer was not found in the data.
     NotFound(NotFoundError),
-    // /// Indicates that a Pointer was malformed.
-    // MalformedPointer(MalformedPointerError),
 }
 
 impl Error {
@@ -40,30 +37,17 @@ impl Error {
     pub fn is_not_found(&self) -> bool {
         matches!(self, Error::NotFound(_))
     }
-    // /// Returns `true` if the error is `Error::MalformedPointerError`.
-    // pub fn is_malformed_pointer(&self) -> bool {
-    //     matches!(self, Error::MalformedPointer(_))
-    // }
 }
-// impl From<MalformedPointerError> for Error {
-//     fn from(err: MalformedPointerError) -> Self {
-//         Error::MalformedPointer(err)
-//     }
-// }
+
 impl From<IndexError> for Error {
     fn from(err: IndexError) -> Self {
         Error::Index(err)
     }
 }
+
 impl From<NotFoundError> for Error {
     fn from(err: NotFoundError) -> Self {
         Error::NotFound(err)
-    }
-}
-
-impl From<OutOfBoundsError> for Error {
-    fn from(err: OutOfBoundsError) -> Self {
-        Error::Index(IndexError::from(err))
     }
 }
 
@@ -79,7 +63,6 @@ impl Display for Error {
             Error::Index(err) => Display::fmt(err, f),
             Error::Unresolvable(err) => Display::fmt(err, f),
             Error::NotFound(err) => Display::fmt(err, f),
-            // Error::MalformedPointer(err) => Display::fmt(err, f),
         }
     }
 }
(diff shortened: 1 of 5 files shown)


asmello:
I'm on the fence now about just using `for_len` as the name for the validating exclusive method. It reads nicely:

```
token.to_index()?.for_len(42)?;
```

Even more so in `Index::Next.for_len(42)`, but I doubt that will come up in real code.

Then we could rename the other methods `for_len_incl` (or `for_len_next`) and `for_len_unchecked`.

chanced:
Ugh, I'm sorry I was slow on the draw on this one. Between getting my daughter a pup and fathers day, I definitely slacked this weekend.

I'm going kayaking with the family today so I won't be available to run through it yet.

I may just hook into this PR and refactor the errors here. It is the last major thing that I want to change about the crate. Dealing with the errors that currently exist is somewhat obnoxious.

chanced:
Nevermind, I'm just going to create a new PR.

Thank you for all of your hard work! I'm sorry the code committed since the start of your participation has been so one sided.

Getting this released will be my top priority going into the week.

asmello:
> Ugh, I'm sorry I was slow on the draw on this one. Between getting my daughter a pup and fathers day, I definitely slacked this weekend.
> 
> I'm going kayaking with the family today so I won't be available to run through it yet.

Don't worry, man. Your family takes priority by a long shot. While I really appreciate your responsiveness, there's no time pressure here, so please pace yourself. Maintaining an open-source project is always a lot of work. I'm glad you're having fun with your daughter and family, I'm a bit jealous of the kayaking.

> Thank you for all of your hard work! I'm sorry the code committed since the start of your participation has been so one sided.

It's cool, I'm just doing these changes because I'm in the mood for it and I happen to have enough time for it right now. We're both volunteers here, you facilitating these changes is more than enough.

By the way, thanks a lot for merging, but did you see my comment about `for_len`? I'm half convinced that's a better naming choice. Happy to push a follow-up PR if you agree.

chanced:
You’re awesome dude. Seriously. 

No, I missed the comment. I’ll be honest, your skillset is well above mine so it takes some time for me to process/understand it all. I’ve been accepting them so not to hold it up and then analyzing it post-merge. I trust that you aren’t malicious and I doubt id catch anything subtle until I'm more well versed in dealing with dynamically sized types.

Re: kayaking. Wife daughter and I picked it up over COVID and absolutely love it. Truly my favorite activity. You should grab yourself a cheap, used sit-on and see if you enjoy it. Gets you out in nature and it's pretty good exercise.

asmello:
> No, I missed the comment. I’ll be honest, your skillset is well above mine so it takes some time for me to process/understand it all. I’ve been accepting them so not to hold it up and then analyzing it post-merge. I trust that you aren’t malicious and I doubt id catch anything subtle until I'm more well versed in dealing with dynamically sized types.

Gotcha. Since you haven't disagreed with the names I mentioned I'll open another PR then. ;)

PR #40, opened 2024-06-17
rename inside -> for_len

asmello:
Follow up of #39. I like this naming convention a bit better.


chanced:
Yea, I think I like `for_len` a bit more than `inside`.

PR #41, opened 2024-06-19
refactoring errors & assignment

chanced:
- Removes `error`, `Error`
- Adds `AssignError`
- Adds `ResolveError`
- Adds `DeleteError`
- Changes `Assign`, `Resolve`, `ResolveMut`, `Delete` traits to now use associated types for `Value` and `Error`
- Removes `MaybePointer`
- Removes `Assignment`
- Adds `Pointer::split_at`
- Changes `Assign::assign` to return the replaced value as `Option<Assign::Value>`


chanced:
This isn't ready yet - i still have a bunch of problems in the tests to fix.

@asmello sorry this just now coming in and still not ready. I know you aren't blocked or in a hurry but I told you I was shooting for a Monday release.

I'm hoping to get it wrapped up either tonight or tomorrow. 

I still need to:
- fix the tests & write a couple more
- Components iter - maybe.
- Refactor `delete` in the same fashion as `assign`

Cargo.toml
```diff
@@ -13,6 +13,7 @@ version       = "0.4.7"
 [dependencies]
 serde      = { version = "1.0", default-features = false, features = ["alloc"] }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
+snafu      = { version = "0.8.3" }
```
asmello, line 16: I assume this choice is because of no-std support?

src/assign.rs
```diff
@@ -1,48 +1,55 @@
-use alloc::borrow::Cow;
-use serde_json::Value;
+use core::mem::replace;
 
-use crate::{Error, Pointer, PointerBuf};
+use serde_json::{map::Entry, Map, Value};
+use snafu::ResultExt;
+
+use crate::{assign_error, AssignError, Pointer, Token};
```
asmello, line 6: We can have a separate discussion about this, but worth converging on a style for imports. I generally prefer to just merge everything into a single block, have never found it useful to separate them, but if you feel strongly against it, I can accept this convention too.

src/assign.rs
```diff
@@ -58,25 +65,176 @@ pub struct Assignment<'a> {
...
+fn assign_to_array<'p, 'd>(
+    token: Token<'_>,
+    remaining: &'p Pointer,
+    dest: &'d mut Vec<Value>,
+    src: Value,
+    offset: usize,
+) -> Result<Assigned<'p, 'd>, AssignError> {
+    let idx = token
+        .to_index()
+        .with_context(|_| assign_error::FailedToParseIndexSnafu { offset })?
```
asmello, line 167: Feels wrong to ignore the original error (here and elsewhere); I think SNAFU supports error sources?

src/pointer.rs
```diff
@@ -1,88 +1,48 @@
...
-fn validate(value: &str) -> Result<&str, MalformedPointerError> {
+fn validate(value: &str) -> Result<&str, ParseError> {
     let mut chars = value.chars();
-
     match chars.next() {
         Some('#') => {
+            // we forgive the usage of `'#'` at the beginning of a pointer even
+            // though this is technically incorrect.
+            // it is a common enough mistake as json pointers are often used in
+            // fragments of URIs to indicate sub-resources.
```
asmello, line 22: Should we take advantage of the major release to remove this relaxation of the spec?

src/pointer.rs
```diff
@@ -1,88 +1,48 @@
...
-                return Err(MalformedPointerError::NoLeadingSlash(value.into()));
-            }
-        }
-        Some('/') => {}
-        Some(_) => {
-            return Err(MalformedPointerError::NoLeadingSlash(value.into()));
+            ensure!(next == Some('/'), parse_error::NoLeadingBackslashSnafu)
         }
+        Some('/') => {}                                          // expected
+        Some(_) => parse_error::NoLeadingBackslashSnafu.fail()?, // invalid pointer - missing leading slash
```
asmello, line 31: Thanks for adding comments here (and other parts of this function)

src/pointer.rs
```diff
@@ -288,6 +248,36 @@ impl Pointer {
             )
             .into()
     }
+    /// Splits the `Pointer` at the given index if the character at the index is
```
asmello, line 251: I wonder if it wouldn't be more useful to index by slash ordinal (rather than character index, which caller may have difficulty choosing?)

src/error.rs
asmello, line 1: Worth noting, I've read in places that SNAFU recommends one error per module.

asmello (commented):
Just did a quick, non-exhaustive look to check the approach and added comments at random places. Will try to do a proper review ASAP, hopefully this Friday, if you have it ready by then.

Cargo.toml
chanced: snafu in general or the version specifically? 

if snafu, I just prefer it over thiserror and definitely favor it over hand-rolled errors. The major driving force for adding it though was `Backtrace`.

If the version, I didn't do an exhaustive check to see which versions are backward compat.

src/assign.rs
chanced: Oh, I have no preference here. I'm lazy and let RA just do its thing. I'm definitely game for cleaning them up. I think there's even a lint that'll do it but its in nightly and was been hit-or-miss for me when I tried it last.

src/assign.rs
chanced: It does and it is. snafu's context system is a bit weird at first but that is wrapping the original error.

The `Snafu` proc macro generates a `Snafu` struct for each variant of an enum error or a single struct for a struct error. The fields on it are anything explicit (backtrace is implicit + [you can create your own](https://docs.rs/snafu/latest/snafu/derive.Snafu.html#controlling-implicitly-generated-data)). Then `context` or `with_context` to inject the source error and implicit data.

shepmaster has written a really good guide here: https://docs.rs/snafu/latest/snafu/guide/index.html

src/pointer.rs
chanced: I'm cool with that. I'll remove it.

src/pointer.rs
chanced: Hmm, yea, that may make sense. 

I added it because I removed owned copies of pointers & tokens from the errors and went with offsets instead. I was debating whether a range of the token instead but just kept it simple.

Hmm.

src/error.rs
chanced: I'm actually going to flatten this all into `lib.rs` on the next PR.

I'm trying to keep it to an error-per fn. Funny enough, shepmaster and I were chatting about last week. Its the same strategy he's finding convenient.

src/pointer.rs
chanced: Heh, I should have had them there from the start :). I plan on adding more and I also need to write out some proper crate-level docs too.

src/pointer.rs
chanced: Either would work post-parsing but we'd still need `offset` for parsing-related errors. Speaking of which, I think I need to add the offset (or whatever we end up using) to `ParseError::InvalidEncoding` as I believe `offset` inside `InvalidEncodingError` is for the `Token`. Or perhaps I just don't re-use `InvalidEncodingError` there.

The only place where it will be a potential point of confusion is parsing if `ParseError` used offsets while `AssignError`, `ResolveError`, and `DeleteError` all used the token index. However, that may still be less confusing overall.

I honestly don't know what the better approach is here.

src/assign.rs
chanced: One last thing on snafu, its [reporting](https://docs.rs/snafu/latest/snafu/attr.report.html) is incredibly nice. It's more for a bin but using snafu at the lib level enables it.

I'm not certain how critical it is to use snafu at the lib level for reporting at the bin level but it's a nice bonus feature.

Cargo.toml
asmello: Not the version, just the choice of crate. snafu is bit heavyweight, but if you deliberately chose it over thiserror, it's fine. I think hand-rolled errors would be fine too, at the scale of complexity of this crate.

src/assign.rs
asmello: I think you're right. It's a bit unfortunate RA doesn't do exactly what I want by default, and there's no setting for it, but if you just merge the blocks once it'll keep them merged (even if you add more imports later). No need to consolidate this everywhere in this PR, but might as well start with it for new code.

src/assign.rs
asmello: > One last thing on snafu, its [reporting](https://docs.rs/snafu/latest/snafu/attr.report.html) is incredibly nice.

That's one thing I actually don't like about `snafu`, it tries to be too many things at the same time. It's a utility for deriving error implementations like `thiserror`, but also bundles its own reporting framework like `anyhow` or `eyre`. I generally think it makes more sense to leave reporting to applications, and have libraries only concern themselves with producing good errors. This separation is a great design decision in the `tracing` crate IMO.

> It does and it is. snafu's context system is a bit weird at first but that is wrapping the original error.

I actually don't like this either 😅 . I don't like implicit magic stuff that gets injected behind the scenes. Generally I believe the intent of code should be understandable without having to look up documentation.

> The Snafu proc macro generates a ___Snafu (where ___ is the name of the error, sans "Error" if it has the suffix) struct for each variant of an enum error or a single struct for a struct error.

This is one I probably like least of all. This naming convention is terrible IMO. In a few cases a suffix like `Error` makes sense to qualify an error type name, when otherwise it would be ambiguous (e.g. `ParseError`), but Snafu doesn't mean anything by itself, and getting it injected unconditionally makes it likely someone will make a compromise, choosing an inferior name for their error just so it doesn't clash with the suffix.

But to be clear, this is all just my personal opinion. I'm sharing as a way to provide a different perspective, but don't consider it part of the code review. I'd never reject a PR based on personal opinion (unless it violated some prior convention).

src/pointer.rs
asmello: Hmm I think there are a couple different cases here. When parsing it makes sense to use `offset` because at that point we still have a string, not a `Pointer`, and the offset is the correct way to point at where the error was originated. The same goes for `Token`, because there it's just another case of parsing.

Once parsing succeeded, though, errors refer to proper `Pointers`, which despite being internally represented as strings are a more structured type. It makes sense for APIs that manipulate pointers to use a different way to refer to positions, especially if they operate on a `Token` basis. I don't think there's any confusion.

Another approach would be to wrap the offset in a newtype, so we can hold the invariant that it points to a valid `/` boundary. That reduces the likelihood of misuse. If this only appears in the context of `Token` manipulation, a token index would be good too, but potentially a direct offset is more efficient in practice, since we can use to skip straight into the position in the internal string.

Yet another option, if the offset is intended so we can find a token in the `Pointer`, is to use a slice instead.

Cargo.toml
chanced: So the way `Backtrace` works is it is (pretty much) zero-cost if you do not have backtraces enabled - they are only captured if enable them via `RUST_BACKTRACE`. If a source error has a backtrace, you can omit it on your own wrapper by including `#[snafu(backtrace)]`.

The major benefit to me was simply that I could have that backtrace available on as many leaf errors as possible as I'm using `snafu` in my work-in-progress json schema crate.

Having said that, other dependencies aren't going to have it, such as`url` and `urn` so it won't be a huge loss if I go back to hand-rolled.

src/assign.rs
chanced: > That's one thing I actually don't like about `snafu`, it tries to be too many things at the same time. It's a utility for deriving error implementations like `thiserror`, but also bundles its own reporting framework like anyhow or eyre.

That's fair. In `snafu`'s defense, the formatting and reporting is only generated in the event you utilize the macro on your main. The formatting itself is fairly minimal simply relies on the appropriate chain of source errors being available to unwrap, for lack of better word.

It does do a lot though, I guess. It sort of merges `thiserror` and `anyhow` through `Whatever`. The idea being, I believe, that both strategies are present to allow folks to gradually step up to proper errors from opaque.

> This is one I probably like least of all. This naming convention is terrible IMO. 

I'll be honest, I'm not a huge fan of it either. If I were to speculate, I'd say he went with the naming convention to ensure he had the least chance of collision on existing artifacts within the source.

It is configurable though via the [`context` attribute](https://docs.rs/snafu/latest/snafu/derive.Snafu.html#controlling-context) for the derive.

> But to be clear, this is all just my personal opinion. I'm sharing as a way to provide a different perspective, but don't consider it part of the code review. I'd never reject a PR based on personal opinion (unless it violated some prior convention).

Oh, no, I know. To be clear, I greatly appreciate your opinions. I very much appreciate your questioning of this. I had (and have) doubts about the added weight & complexity before you voiced your concerns.

src/pointer.rs
chanced: > Once parsing succeeded, though, errors refer to proper Pointers, which despite being internally represented as strings are a more structured type. It makes sense for APIs that manipulate pointers to use a different way to refer to positions, especially if they operate on a Token basis. I don't think there's any confusion.

That's fair, totally on board with that line of thought.

> Another approach would be to wrap the offset in a newtype, so we can hold the invariant that it points to a valid / boundary. 

I like this approach but we'd need a set of methods to facilitate it so there is utility beyond error location -> token. It'd need to become a construct outside of that flow or it'll be an odd appendage.

I think whatever it is we decide, we need a way to expose it via an iterator. We are currently looping through tokens using your awesome `split_front` and `split_back` methods, which is a hell of a lot better than it was before (with all the allocations and such) but it feels like we could have a more purpose-built solution to needing to walk a pointer forward/backward.

> Yet another option, if the offset is intended so we can find a token in the Pointer, is to use a slice instead.

Sorry, I don't understand the option on this one.

src/assign.rs
chanced: > I actually don't like this either 😅 . I don't like implicit magic stuff that gets injected behind the scenes. Generally I believe the intent of code should be understandable without having to look up documentation.

I completely get where you're coming from. I'm generally avidly opposed to magic. For example, a lot of people _love_ rocket but it is far too 🪄 ✨ for me. At first glance, `snafu` seemed the same. It took me using it to finally come around and appreciate the ergonomics of it.

Having said all that, I think I'll remove it and go back to hand-rolled errors. This is a small, primitive crate that probably shouldn't burden the ecosystem with unnecessary dependencies. While `Backtrace` is nice, its probably more of a bin concern or, at min, a consuming lib.

src/assign.rs
chanced: I think the untidiness of imports is amplified by all the unnecessary mods in the crate. Flattening it should make this a lot easier to keep in check. 

I may try that lint again too. I was super excited when I found it because while I do find the utility of auto-imports super convenient, I do get annoyed and distracted by cluttered use blocks.

src/pointer.rs
chanced: https://github.com/chanced/jsonptr/issues/42

chanced:
I went ahead and backed out `snafu` for hand-rolled errors. I'm leaving `offset` intact for now until we come up with the appropriate replacement.

chanced:
I'm going to return a `PointerBuf` as part of `Asssignment` but it could be `Cow<'p, Pointer>` so that for most cases, it'd have the lifetime of the input `Pointer` but in the case of one that included `"-"`, it'd be `PointerBuf`.

I dont know if its worth it. On one hand, more logic. On the other, i doubt `"-"` will show up much so most could be borrowed.


i'm running out of fumes tonight but I'll circle back to this in the morning.

chanced:
@asmello hey, one of the quickcheck tests (`pointer::tests::qc_intersection`) is failing but i dont see anything obvious.

I'm not sure how I'm tripping it.

chanced:
@asmello this method on `Pointer` is no longer used. I'm just checking to see if you have utility for it before I delete it:

```rust
    fn partial_path(&self, suffix: &Self) -> &Self {
        self.strip_suffix(suffix).expect("suffix came from self")
    }
```

chanced:
I think I know what it is.

chanced:
Added more tests for `parse`, some of which are failing.

chanced:
The only thing left that I wanted to do was add the `Components` iterator. It may be the best place to include the index, if we go with something other than offsets.

src/assign.rs
asmello: Yeah, didn't want to push hard against it, but my gut feeling is that unless one is building a complex crate that already has a ton of dependencies, staying lean is best. It really doesn't matter most of the time for consumers, but when it does it's a lot of pain. I'm less concerned about the choice of `snafu` vs `thiserror` than about that.

src/pointer.rs
asmello: Yeah that makes sense. My main worry with using token indices is the need to traverse the string to count them — which is perfectly reasonable to have in the API, but if we use it internally the cost compounds. I like the idea of doing this via an iterator. Probably makes it easier to change the implementation later if necessary.

chanced:
@asmello if you're good with this so far, I'll merge it in. I don't want to cut a release until we shake out all of the remaining breaking changes but this gets the logic mostly where it needs to be. There are some minor things, like restructuring the mods, `Pointerbuf::Append` accepting `&Pointer` or `T: AsRef<Pointer>`, and whether or not we change offsets to indexes.

asmello:
Still catching up, think I'll need to do a proper review tomorrow as it's already getting late here. But I'll comment here and there.

> for mod structure, I'm not sure which would be better

I'm not sure there are any official guidelines for this. There seems to be a bit of a philosophical disagreement, though there are some common points. I think we should avoid exporting the same symbol in more than one path (which I think I may have been guilty of with `Index` - let's fix that). Other than that, a good rule of thumb is probably to export the most commonly used symbols at the top-level, and leave the less used ones (and in particular the ones that users can't instantiate directly) accessible in nested modules.

What constitutes "commonly used symbols" is open to interpretation. Having used `jsonptr` a fair bit, I'm confident we can include `Pointer`, `PointerBuf`, `Token` and probably `Error` in that category. `Index` probably shouldn't, as it's primarily used anonymously.

I haven't really used the the assign/resolve/delete API much in practice, so I don't have much of an opinion on it, but it does seem to me like a less core component of the crate. May seem surprising, since the README examples focus on those APIs heavily, but beyond my own experience, the fact this API is not used in `json_patch` adds some weight to this view. Not to say we should remove that code, but perhaps it can be under a feature flag. Regardless, I think those symbols shouldn't be at the top level. They should probably still, however, be fairly high up. I think it'd be reasonable to export them all in common module. Not sure what to name it - maybe `evaluation`? A bit long though.

asmello:
> @asmello if you're good with this so far, I'll merge it in. I don't want to cut a release until we shake out all of the remaining breaking changes but this gets the logic mostly where it needs to be. There are some minor things, like restructuring the mods, `Pointerbuf::Append` accepting `&Pointer` or `T: AsRef<Pointer>`, and whether or not we change offsets to indexes.

If this is holding up other changes feel free to merge and I can review async. I agree with the general approach from what I skimmed over.

chanced:
> Still catching up, think I'll need to do a proper review tomorrow as it's already getting late here. But I'll comment here and there.

Sounds good. I'd rather wait until you've had a chance to run through it.

asmello:
> edit: fixed it, sorry about the mention. I wasn't sure why it randomly tripped and stayed consistent. I haven't used quickcheck before so it tripped me up.

Yeah no worries. I've had a very busy week, unfortunately, so apologies for not engaging much so far. Glad you figured out the issue. Quickcheck is a powerful tool to automate some edge case testing, but it takes some getting used to for interpretation.

asmello:
> @asmello this method on `Pointer` is no longer used. I'm just checking to see if you have utility for it before I delete it:
> 
> ```rust
>     fn partial_path(&self, suffix: &Self) -> &Self {
>         self.strip_suffix(suffix).expect("suffix came from self")
>     }
> ```

It's a private method, if it's no longer used, drop it ruthlessly. 😄

chanced:
> so apologies for not engaging much so far
Oh no need to apologize whatsoever! I feel bad that I said I was hoping to have this done on Monday and it is now Friday. It's been a week on this side too.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+    /// An `Index` `Token` within the `Pointer` was out of bounds of the
+    /// target array.
+    OutOfBounds {
+        /// Offset of the partial pointer starting with the invalid index.
+        offset: usize,
+        /// The source [`OutOfBoundsError`]
+        source: OutOfBoundsError,
+    },
+}
+impl fmt::Display for AssignError {
```
asmello, line 137:
```suggestion
}

impl fmt::Display for AssignError {
```

asmello, line 137:
```suggestion
}

impl fmt::Display for AssignError {
```

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+    let idx = token
+        .to_index()
+        .map_err(|source| AssignError::FailedToParseIndex { offset, source })?
+        .for_len_incl(array.len())
+        .map_err(|source| AssignError::OutOfBounds { offset, source })?;
+    buf.push_back(idx.into());
+
+    debug_assert!(idx <= array.len());
+
+    if idx < array.len() {
```
asmello, line 186: Could also take advantage of the fact that `Index` is an enum and branch here on the variant.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+                next_buf: buf,
+            })
+        }
+    } else {
+        // element does not exist in the array.
+        // we create the path and assign the value
+        let src = expand_src_path(remaining, src)?;
+        array.push(src);
+        // SAFETY: just pushed.
+        let assigned = array.last_mut().unwrap();
```
asmello, line 213:
```suggestion
        let assigned = array.last_mut().expect("just pushed");
```
`SAFETY` comments are reserved for unsafe code. Being consistent about this helps tooling check that all unsafe code has been accounted for. Plus using `expect` is generally good practice because it gives context in case the expectation is violated.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+        assigned: value,
+        assigned_to: buf,
+        replaced,
+    }))
+}
+
+fn expand_src_path(mut path: &Pointer, mut src: Value) -> Result<Value, AssignError> {
+    while let Some((ptr, tok)) = path.split_back() {
+        path = ptr;
+        match tok.decoded().as_ref() {
```
asmello, line 296: This can be `encoded`, since the matched characters won't ever need decoding. Probably worth a note stating that observation, though.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+    fn test_assign() {
+        let mut data = json!({});
+        let ptr = Pointer::from_static("/foo");
+        let val = json!("bar");
+
+        let assignment = ptr.assign(&mut data, val.clone()).unwrap();
+        assert_eq!(assignment.replaced, None);
+        assert_eq!(assignment.assigned, &val);
+        assert_eq!(assignment.assigned_to, "/foo");
+
```
asmello, line 326: Probably worth adding a check on `data` directly, too.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
-    /// use serde_json::json;
-    /// use jsonptr::{Pointer, Assign};
+    /// # use serde_json::json;
+    /// # use jsonptr::{Pointer, Assign};
     /// let mut data = json!({ "foo": ["zero"] });
     /// let mut ptr = Pointer::from_static("/foo/-");
     /// let assignment = data.assign(&mut ptr, "one").unwrap();
     /// assert_eq!(assignment.assigned_to, Pointer::from_static("/foo/1"));
     /// ```
     pub assigned_to: PointerBuf,
```
asmello, line 69: I don't think this is worth changing in this MR, but we should consider if the cost of keeping track of the assignment path is worth it, since it adds extra allocations. Could maybe be an opt-in thing, either through a flag or separate method.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+}
+
+#[cfg(test)]
+mod tests {
+    use serde_json::{json, Value};
+
+    use crate::{Pointer, PointerBuf};
+
+    #[test]
+    fn test_assign() {
```
asmello, line 317: Probably also worth testing assigning directly to root (replacing it), both from an empty base object and a non-empty one.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+        // now testing replacement
+        let val2 = json!("baz");
+        let assignment = ptr.assign(&mut data, val2.clone()).unwrap();
+        assert_eq!(assignment.replaced, Some(Value::String("bar".to_string())));
+        assert_eq!(assignment.assigned, &val2);
+        assert_eq!(assignment.assigned_to, "/foo");
+    }
+
+    #[test]
+    fn test_assign_with_explicit_array_path() {
```
asmello, line 336: Worth also checking what happens if:
1. The array exists, but is empty
2. The array has one element
3. The array has two elements
4. The path ends at the index

Could potentially cover this with a quickcheck test though.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+                obj.insert(tok.to_string(), src);
+                src = Value::Object(obj);
+            }
+        }
+    }
+    Ok(src)
+}
+
+#[cfg(test)]
+mod tests {
```
asmello, line 311: Worth adding some unhappy path tests too.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+            Ok(Assigned::Continue {
+                next_value: &mut array[idx],
+                same_src: src,
+                next_buf: buf,
+            })
+        }
+    } else {
+        // element does not exist in the array.
+        // we create the path and assign the value
+        let src = expand_src_path(remaining, src)?;
```
asmello, line 210: Hmm I think there's an issue here if the token contains a non-zero number - it'll create a map in that case.

asmello (commented):
Ok, was only able to review one file so far. Will continue tomorrow.

src/index.rs
```diff
@@ -153,11 +153,7 @@ impl TryFrom<&Token<'_>> for Index {
         if value.encoded() == "-" {
             Ok(Index::Next)
         } else {
-            value
-                .decoded()
-                .parse::<usize>()
-                .map(Index::Num)
-                .map_err(ParseIndexError::new)
+            Ok(value.decoded().parse::<usize>().map(Index::Num)?)
```
asmello, line 156: Oh nice!

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
 mod delete;
-pub use delete::*;
+pub use delete::Delete;
 mod resolve;
-pub use resolve::*;
+pub use resolve::{Resolve, ResolveError, ResolveMut};
 
 pub mod prelude;
 
+// TODO(chanced): move these into lib.rs
```
asmello, line 18: Wait, not sure I follow this. Exporting symbols from these in `lib.rs` is fine, but I think it's an anti-pattern to dump everything public here. If you are trying to get rid of the glob uses, I think it's best to just enumerate the symbols as you've done above.

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+mod pointer;
+pub use pointer::*;
+mod token;
+pub use token::*;
+pub mod index;
+pub use index::Index;
 
 #[cfg(test)]
 mod arbitrary;
+
```
asmello, line 30: Also not sure I get the rationale for moving these out of a dedicate error module?

chanced:
> I haven't really used the the assign/resolve/delete API much in practice,

Yea, they way they are currently built just aren't useful at all. I have been chipping away at refactoring them in this PR but I've been doubting their usefulness. I have partially included a refactor in the last push. It isn't complete (`AssignValue` would need to be generic, and other traits would need to be updated).

The update would allow implementations for `yaml`, `toml` (through feature flags),or whatever else someone wanted to impl.

I'm actually cool with not doing this though and simply always accepting `serde_json::Value`

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+/// Indicates that a `Pointer` was malformed and unable to be parsed.
+#[derive(Debug, PartialEq)]
+pub enum ParseError {
+    /// `Pointer` did not start with a backslash (`'/'`).
+    NoLeadingBackslash,
+
+    /// `Pointer` contained invalid encoding (e.g. `~` not followed by `0` or
+    /// `1`).
+    InvalidEncoding {
```
asmello, line 39: One caveat to take into account for struct-like variants like this is that you're committing to the fields signature being part of the public interface. Probably ok here, I think.

src/lib.rs
chanced: Oh, those were from before the comment above about restructuring the surface.

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+impl fmt::Display for ParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::NoLeadingBackslash { .. } => {
+                write!(
+                    f,
+                    "json pointer is malformed as it does not start with a backslash ('/')"
+                )
+            }
+            Self::InvalidEncoding { source, .. } => write!(f, "{}", source),
```
asmello, line 57:
```suggestion
            Self::InvalidEncoding { source, .. } => write!(f, "{source}"),
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+        }
+    }
+}
+impl ParseError {
```
asmello, line 61:
```suggestion
}

impl ParseError {
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+    /// Returns `true` if this error is `NoLeadingBackslash`; otherwise returns
+    /// `false`.
+    pub fn is_no_leading_backslash(&self) -> bool {
+        matches!(self, Self::NoLeadingBackslash { .. })
+    }
+    /// Returns `true` if this error is `InvalidEncoding`; otherwise returns
```
asmello, line 67:
```suggestion
    }
    
    /// Returns `true` if this error is `InvalidEncoding`; otherwise returns
```

src/lib.rs
chanced: it was mostly for convenience with the expectation that the surface was going to change. I wasn't sure where everything would end up but I don't think `error` needs a mod to itself. I'm guessing the 2 error types (`AssignError`, `ResolveError`) will get moved, either into `lib.rs` but if `assign`, `resolve`, and `delete` stay as mods, they make sense to live alongside their impl.

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+    }
+}
+
+/// Indicates that the `Token` could not be parsed as valid RFC 6901 index.
+#[derive(Debug, PartialEq, Eq)]
+pub struct ParseIndexError {
+    /// The source `ParseIntError`
+    pub source: ParseIntError,
+}
+impl From<ParseIntError> for ParseIndexError {
```
asmello, line 143:
```suggestion
}

impl From<ParseIntError> for ParseIndexError {
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+    fn from(source: ParseIntError) -> Self {
+        Self { source }
+    }
+}
+impl fmt::Display for ParseIndexError {
```
asmello, line 148:
```suggestion
}

impl fmt::Display for ParseIndexError {
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+
+/// A token within a json pointer contained invalid encoding (`~` not followed
+/// by `0` or `1`).
+///
+#[derive(Debug, PartialEq, Eq)]
+pub struct InvalidEncodingError {
+    /// offset of the erroneous `~` from within the `Token`
+    pub offset: usize,
+}
+impl InvalidEncodingError {
```
asmello, line 169:
```suggestion
}

impl InvalidEncodingError {
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+    /// The byte offset of the first invalid `~`.
+    pub fn offset(&self) -> usize {
+        self.offset
+    }
+}
+impl fmt::Display for InvalidEncodingError {
```
asmello, line 175:
```suggestion
}

impl fmt::Display for InvalidEncodingError {
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+/// Returned from `Pointer::replace_token` when the provided index is out of
+/// bounds.
+#[derive(Debug, PartialEq, Eq)]
+pub struct ReplaceTokenError {
+    /// The index of the token that was out of bounds.
+    pub index: usize,
+    /// The number of tokens in the `Pointer`.
+    pub count: usize,
+}
+impl fmt::Display for ReplaceTokenError {
```
asmello, line 259:
```suggestion
}

impl fmt::Display for ReplaceTokenError {
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+    pub offset: usize,
+}
+
+impl fmt::Display for NotFoundError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "path starting at offset {} not found", self.offset)
+    }
+}
+#[cfg(feature = "std")]
+impl std::error::Error for NotFoundError {}
```
asmello, line 248:
```suggestion

#[cfg(feature = "std")]
impl std::error::Error for NotFoundError {}
```

src/lib.rs
```diff
@@ -3,27 +3,264 @@
...
+impl fmt::Display for OutOfBoundsError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "index {} out of bounds (limit: {})",
+            self.index, self.length
+        )
+    }
+}
+#[cfg(feature = "std")]
```
asmello, line 231:
```suggestion
}

#[cfg(feature = "std")]
```

chanced:
@asmello dude, go get some sleep. This can wait.

asmello:
> > I haven't really used the the assign/resolve/delete API much in practice,
> 
> Yea, the way they are currently built just aren't useful at all. I have been chipping away at refactoring them in this PR but I've been doubting their usefulness. I have partially included a refactor in the last push. It isn't complete (`AssignValue` would need to be generic, and other traits would need to be updated).
> 
> The update would allow implementations for `yaml`, `toml` (through feature flags),or whatever else someone wanted to impl.
> 
> I'm actually cool with not doing this though and simply always accepting `serde_json::Value` and deleting the traits.

Oh I think this is all useful, especially post-refactor, what I mean is more that the basic `Pointer` type and companions is already useful on its own, so some users may want to consume just that part in isolation.

src/lib.rs
asmello: It's not terrible to include these in the root. I do think an `error` module makes sense if you have global error types, but I guess you got rid of those. In that case yeah, defining them close to where they are produced is the best, lacking that here is an ok fallback.

src/assign.rs
chanced: I think that's okay, sort of. Any non `"0"` index is going to be turned into an object too. 
e.g., this is valid:
```json
{ "-": "valid json", "0": "still valid" }

I've been kind of fretting over it because an intended object with keys `"-"` or `"0"` is going to spit out an array. So if the keys are dynamic and someone uses one of those as keys and assumptions are made 😭 . There just really isn't any clean way around it though and its the only way to properly allow for auto-expansion.

Perhaps that's the answer. I'm not sure.

File: src/assign.rs (hunk 15 above)
Comment (chanced) [reply src/assign.rs:313]: The first one is but I'll add more. Never can have too many.

```rust
            (
                json!({"foo": "bar"}), // data
                "/foo/bar",            // ptr
                json!({"foo": "bar"}), // expected_data
                None,                  // expected_deleted
            ),
```
`delete` doesn't error out, so `None` is basically did not exist.

chanced: woops, I'm sorry. I clicked through and i somehow ended up on deletes test. Yea, I'll add more unhappy path assigns.

src/assign.rs
asmello: Hmm yeah, we've faced the same problem when using `jsonptr` to access a different JSON-like data structure. A possible solution here could be to allow the caller to choose a strategy, and a good default is probably to just not expand at all (other than when it's an unambiguous append-to-array situation).

chanced: I considered a strategy, but you'd either need to opt in to all objects or all possible arrays. Only the first option would prevent accidentally getting the wrong shape of data. I guess the `Strategy` could be something to the effect of
```rust
pub enum Expansion {
    Auto,
    Error
}
```

chanced:
> Oh I think this is all useful, especially post-refactor, what I mean is more that the basic `Pointer` type and companions is already useful on its own, so some users may want to consume just that part in isolation.

Oh, I see. I thought you meant the traits. They were mostly useless before. I'll finish this refactor then. We can decide whether to dump them or go with `serde_json::Value`.

I like the idea of making `assign`, `resolve`, and `delete` feature flags that are enabled by default.

src/assign.rs
chanced: So I fixed `assigned_to` in this PR. It now reports appropriately the path of `assigned` (mut ref to the src assigned). Before, it was only being updated if `"-"` existed in the path, basically being the full pointer. Now it's wherever the assignment ends.

If we keep auto-expansion of any sort, this probably needs to stay. However, it could be made into a `Cow<'p, Pointer>`. It'll take more work / logic but in most cases, it won't allocate.

src/assign.rs
```diff
@@ -58,25 +54,456 @@ pub struct Assignment<'a> {
...
+    pub replaced: Option<Value>,
+}
+
+pub(crate) fn assign_value<'v>(
+    mut ptr: &Pointer,
+    mut value: &'v mut Value,
+    mut src: Value,
+) -> Result<Assignment<'v>, AssignError> {
+    let mut offset = 0;
+    let mut buf = PointerBuf::with_capacity(ptr.as_str().len());
```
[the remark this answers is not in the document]
chanced: Oh, thanks for mentioning that. I had an error (using `str::len` rather than `Chars::len`).

This does bring up a question as to what should `len` report. `Token` count?, byte count like `str`, or `Chars::len`

src/assign.rs
chanced: awesome, thanks

src/assign.rs
chanced: this is so `assign_value` can continue the traversal.

chanced:
I drafted a potential solution to expansion in `Assign`

chanced:
I'm not super excited about the name of `AutoExpand` or the variants (`Enabled`, `Disabled`) but they work for on/off. If more are added, such as maybe `"AllObjects"` that always creates object paths irregardless of the token, then they'd need proper names.

I also need to revisit what is passed in. This was just the min to get it working. Current pointer (path up to the token) should probably be provided as well. If so, a `struct` may be needed as two `&Pointer`s as parameters may yield ambiguity as to which is which.

asmello:
> I'm not super excited about the name of AutoExpand or the variants (Enabled, Disabled) but they work for on/off. If more are added, such as maybe "AllObjects" that always creates object paths irregardless of the token, then they'd need proper names.

`Disabled` is probably fine, don't think it is invalidated if more are added, but how about `Default` for the other? Could always also tweak the meaning without it being a semantic break.

> If so, a struct may be needed as two &Pointers as parameters may yield ambiguity as to which is which.

Sounds reasonable!

General comment: I think things are still a bit in flux (especially given the `todo!()` you added as part of the strategy refactor), but don't forget to run clippy before you're done, there's a few warnings I won't comment about. ;)

src/pointer.rs
```diff
@@ -288,6 +248,36 @@ impl Pointer {
...
+    /// assert_eq!(head, Pointer::from_static("/foo"));
+    /// assert_eq!(tail, Pointer::from_static("/bar/baz"));
+    ///
+    /// assert_eq!(ptr.split_at(3), None);
+    /// ```
+    pub fn split_at(&self, idx: usize) -> Option<(&Self, &Self)> {
+        if self.0.get(idx..idx + 1) != Some("/") {
+            return None;
+        }
+        let (head, tail) = self.0.split_at(idx);
```
asmello, line 278: Yeah I think I can do the same for `split_front` and `split_back` not allocate. Originally they did because `Token` was owned, but now that it isn't I think it should be possible to mimic std.

src/pointer.rs
```diff
@@ -469,110 +400,28 @@ impl Pointer {
...
-            Ok(Assignment {
-                assigned: Cow::Borrowed(dest),
-                replaced,
-                created_or_mutated: created_or_mutated.unwrap_or_else(|| assigned_to.clone()),
-                // NOTE: in practice, `assigned_to` is always equivalent to the original pointer,
-                // with `-` replaced by the actual index (if it occurs)
-                assigned_to,
-            })
-        }
+        dest.assign(self, src, expand_strategy)
```
asmello, line 424: Nice to move all of these elsewhere. :+1:

src/pointer.rs
```diff
@@ -469,110 +400,28 @@ impl Pointer {
     ///
     /// ## Example
     /// ```rust
-    /// use jsonptr::Pointer;
+    /// use jsonptr::{ Pointer, AutoExpand };
```
asmello, line 403: Tiny nit: extra spaces.

src/pointer.rs
```diff
@@ -710,6 +548,15 @@ impl PointerBuf {
         Self(String::new())
     }
 
+    /// Attempts to parse a string into a `PointerBuf`.
+    pub fn parse(s: &str) -> Result<Self, ParseError> {
```
asmello, line 552: I suppose it makes sense to provide a shortcut for this. :+1:

src/pointer.rs
```diff
@@ -710,6 +548,15 @@ impl PointerBuf {
...
+        Pointer::parse(s).map(Pointer::to_buf)
+    }
+
+    /// Creates a new `PointerBuf` with the given capacity.
+    pub(crate) fn with_capacity(capacity: usize) -> Self {
+        Self(String::with_capacity(capacity))
+    }
     /// Creates a new `PointerBuf` from a slice of non-encoded strings.
```
asmello, line 560:
```suggestion
    }

    /// Creates a new `PointerBuf` from a slice of non-encoded strings.
```

asmello:
Done with a proper review! Generally LGTM, solid code! Mostly minor stuff. The amount of tests gives me confidence there's nothing too bad I may have missed. :joy:

src/assign.rs
```diff
@@ -1,48 +1,93 @@
...
 pub trait Assign {
+    /// The type of value that is being assigned.
+    ///
+    /// Provided implementations include:
+    ///
+    /// | Lang  |     value type      | feature flag |
+    /// | ----- | ------------------- |: ---------- :|
+    /// | JSON  | `serde_json::Value` |   `"json"`   |
+    /// | YAML  | `serde_json::Value` |   `"yaml"`   |
+    /// | TOML  | `serde_json::Value` |   `"toml"`   |
```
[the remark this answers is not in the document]
chanced: woops, good catch :)

chanced:
> General comment: I think things are still a bit in flux (especially given the todo!() you added as part of the strategy refactor), but don't forget to run clippy before you're done, there's a few warnings I won't comment about. ;)

Yea, I didn't have time to flush it out but wanted to get it pushed incase you had a chance to look.

src/token.rs
```diff
@@ -49,11 +49,10 @@ impl<'a> Token<'a> {
     /// assert_eq!(err.offset(), 3);
     /// ```
     pub fn from_encoded(s: &'a str) -> Result<Self, InvalidEncodingError> {
-        let err_at = |i| Err(InvalidEncodingError::new(i));
```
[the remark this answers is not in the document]
chanced: Yep, I saw that. Its a clean way of avoiding it. It just didn't end being needed once the errors were refactored.

src/assign.rs
chanced: It'd be nice if rustfmt's [`blank_lines_lower_bound`](https://rust-lang.github.io/rustfmt/?version=v1.4.30&search=#blank_lines_lower_bound) had a configuration that didn't double space every line but instead just the end of blocks.

src/assign.rs
```diff
@@ -1,48 +1,93 @@
...
+    fn assign<'v, V, E>(
+        &'v mut self,
+        ptr: &Pointer,
+        value: V,
+        expand_strategy: E,
+    ) -> Result<Assignment<'v, Self::Value>, Self::Error>
     where
-        V: Into<Value>;
+        V: Into<Self::Value>,
+        E: Expand<Self::Value>;
```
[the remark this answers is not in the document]
chanced: > is the reason for this being a trait that you expect users to bring their own strategy?

Yea, the idea was that we'd provide a few options but users could use their own logic for deciding how to structure the expansion.

Cargo.toml
```diff
@@ -11,12 +11,21 @@ repository    = "https://github.com/chanced/jsonptr"
...
+serde_json = { version = "1.0", optional = true, default-features = false, features = [
+	"alloc",
+] }
+serde_yaml = { version = "0.9.34", optional = true }
+toml = { version = "0.8", optional = true }
 
 [features]
-default = ["std"]
+assign  = []
+default = ["std", "json", "resolve", "assign", "delete"]
```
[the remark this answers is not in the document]
chanced: I'm only using `Resolve` in the json schema crate and I suspect that only needing `resolve` would be more common than needing `assign` and/or `replace`. 

Agreed that its a small crate though so this may not be worth the flag at all. Hmm.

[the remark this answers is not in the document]
asmello: True... I think that'd be reasonable too. That said, I think it makes sense to split-off resolve if you do end up adding flags.

src/assign.rs
```diff
@@ -58,25 +103,477 @@ pub struct Assignment<'a> {
...
+                AutoExpand::Enabled => expand_auto(path, src),
+                AutoExpand::Disabled => expand_error(path, src),
+            }
+        }
+    }
+
+    fn expand_auto(mut path: &Pointer, mut src: Value) -> Result<Value, AssignError> {
+        while let Some((ptr, tok)) = path.split_back() {
+            path = ptr;
+            match tok.decoded().as_ref() {
```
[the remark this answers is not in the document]
chanced: not sure what this one is pointing at

src/assign.rs
```diff
@@ -1,48 +1,93 @@
...
+    Enabled,
+
+    /// This strategy will error if the [`Pointer`] is not fully exhausted before
+    /// a scalar or non-existent key or [`Index`](crate::Index) is encountered.
+    Disabled,
+}
+
+impl Default for AutoExpand {
+    fn default() -> Self {
+        Self::Enabled
```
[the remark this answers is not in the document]
chanced: Good point!

src/resolve.rs
```diff
@@ -1,30 +1,417 @@
...
 pub trait Resolve {
+    /// The type of value that is being resolved.
+    ///
+    /// Provided implementations include:
+    ///
+    /// | Lang  |     value type      | feature flag |
+    /// | ----- | ------------------- |: ---------- :|
+    /// | JSON  | `serde_json::Value` |   `"json"`   |
+    /// | YAML  | `serde_json::Value` |   `"yaml"`   |
+    /// | TOML  | `serde_json::Value` |   `"toml"`   |
```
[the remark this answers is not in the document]
chanced: thanks, copy-paste fail.

src/resolve.rs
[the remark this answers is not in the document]
chanced: absolutely, i wish there was a lint that just added newlines after blocks.

src/assign.rs
```diff
@@ -58,25 +103,477 @@ pub struct Assignment<'a> {
...
+        mut ptr: &Pointer,
+        mut value: &'v mut Value,
+        mut src: Value,
+        expand_strat: E,
+    ) -> Result<Assignment<'v, Value>, AssignError> {
+        let mut offset = 0;
+        let mut buf = PointerBuf::with_capacity(ptr.as_str().len());
+
+        while let Some((token, tail)) = ptr.split_front() {
+            let tok_len = token.encoded().chars().count();
```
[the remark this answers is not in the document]
chanced: not sure what this one is pointing at. `let tok_len = token.encoded().chars().count();` maybe?

src/assign.rs
chanced: I haven't used this API either but it seems like someone may want to know what ended up getting assigned and where. I may switch it over to `Cow<'p, Pointer>` even if it is always `Cow::Owned` for now.

src/resolve.rs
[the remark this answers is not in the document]
chanced: I'll do better with spacing :)

src/delete.rs
```diff
@@ -1,35 +1,120 @@
...
+                    Some(json!("qux")),
+                ),
+            ];
+            for (mut data, ptr, expected_data, expected_deleted) in tests {
+                let ptr = Pointer::from_static(ptr);
+                let deleted = ptr.delete(&mut data);
+                assert_eq!(
+                    expected_data,
+                    data,
+                    "\ndata not as expected
```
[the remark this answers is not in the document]
chanced: ooooh, that's cool!

src/pointer.rs
```diff
@@ -175,10 +131,14 @@ impl Pointer {
     /// Attempts to parse a string into a `Pointer`.
     ///
     /// If successful, this does not allocate.
-    pub fn parse<S: AsRef<str> + ?Sized>(s: &S) -> Result<&Self, MalformedPointerError> {
+    pub fn parse<S: AsRef<str> + ?Sized>(s: &S) -> Result<&Self, ParseError> {
         validate(s.as_ref()).map(Self::new)
     }
 
+    pub fn len(&self) -> usize {
```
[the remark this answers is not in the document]
chanced: ah! I didn't realize I had! I'm not sure its a good idea though. Hmm.

src/pointer.rs
```diff
@@ -1,91 +1,47 @@
...
         if c == '~' {
-            match chars.next() {
-                Some('0') | Some('1') => {}
-                _ => {
-                    return Err(MalformedPointerError::InvalidEncoding(value.into()));
-                }
+            // pulling down
+            let next = chars.next().map(|(_, c)| c);
+            if !matches!(next, Some('0') | Some('1')) {
+                println!("next: {:?}", next);
```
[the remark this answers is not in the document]
chanced: woo, thank you! I completely forgot that it was there.

src/assign.rs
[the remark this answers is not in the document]
asmello: This doesn't need to call `decoded`, it's sufficient to use `encoded` since the string it's comparing to is never one of the encoded symbols.

src/assign.rs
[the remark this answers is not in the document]
asmello: Yes... Does it not render the line it's anchored to for you?

src/resolve.rs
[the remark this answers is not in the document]
asmello: No worries, man, it's a minor thing and everyone does this. There really should be a rustfmt rule for it.

src/pointer.rs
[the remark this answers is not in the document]
asmello: I think it's fine. There's a case to be made that there's some ambiguity around encoding (should `~0` count as one character or two?) but the same applies to `str` itself (a character may use up multiple bytes), so I think it's fine to follow std. Should definitely add docs though, on a second look.

src/assign.rs
```diff
@@ -1,48 +1,93 @@
...
+    ///   length, pushing onto the array.
+    ///  
+    ///  Note: This strategy will not return [`AssignError::NotFound`] or
+    ///  [`AssignError::Unreachable`].
+    ///
+    Enabled,
+
+    /// This strategy will error if the [`Pointer`] is not fully exhausted before
+    /// a scalar or non-existent key or [`Index`](crate::Index) is encountered.
+    Disabled,
```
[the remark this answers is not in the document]
chanced: Thanks :)

src/pointer.rs
chanced: I think we are good - there are so few allocations on this. Your changes have been such a huge improvement on that already!

Fixing this crate was going to be next on my list after my json schema crate. I'm so grateful I don't have to do that now, especially since this turned out way better than what I would have produced!

Cargo.toml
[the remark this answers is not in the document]
chanced: I went ahead and made it granular with an `"ops"` flag that enables all. It may be a bit overkill but it should be fine.

src/assign.rs
chanced: im removing it. its akward when it comes to replacement of key,

```rust
let data = json!{ "foo": "str"};
let assignment = data.assign(Pointer::from_static("/foo/bar/baz", json!("qux")).unwrap();
// should this be assigned to `"foo"` or `"foo/bar"`? 
assert_eq!(assignment.assigned_to, ???);
assert_eq!(assignment.assigned, json!({"bar": {"baz": "qux"}}))
```

src/assign.rs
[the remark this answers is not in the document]
chanced: No, it's throwing it all over the place for some reason. I think its once they become outdated and the code gets shifted? It's weird that it doesn't re-calibrate offsets.

src/assign.rs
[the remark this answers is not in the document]
chanced: oooh! good catch!

src/lib.rs
chanced: oh, I hadn't considered that. I think I'm okay with it in this case. Any changes to the error are going to come with other breaking changes. But yea, I'll definitely keep that in mind in the future.

Thanks.

chanced:
@asmello The more I worked on `Assign`, the more apparent it became that trying to provide a strategy for replacement seemed like it wasn't the right solution. Also returning a mutable reference became a hassle.

I walked it back and deleted `Assignment` in favor of returning `Option<Value>` (replaced).

The possibility of ambiguity is too high for this to be used in anything that demands precision. I think there's a better solution for more controlled assignment, perhaps something like an iterator of entries. I'm not quite sure yet.

Either way, I don't think we need to solve it now. At least not for this release.

Let me know what you think. I'll work on adding tests and documentation in the meantime. I'll merge this in if you're cool with abandoning the effort.

asmello:
> @asmello The more I worked on `Assign`, the more apparent it became that trying to provide a strategy for replacement just wasn't the right solution. Also dealing with the return of a mutable reference became a hassle.
> 
> I walked it back and deleted `Assignment` in favor of returning `Option<Value>` (replaced).
> 
> The possibility of ambiguity is too high for this to be used in anything that demands precision. I think there's a better solution for more controlled assignment, perhaps something like an iterator of entries. I'm not quite sure yet.
> 
> Either way, I don't think we need to solve it now. Neither of us are using this API and we haven't received feedback from anyone who is. I think it's best just to keep it simple with a helpful API that is only suitable for certain situations.
> 
> Let me know what you think. I'll work on adding tests and documentation in the meantime. I'll merge this in if you're cool with abandoning the effort.

That works for me. I'm always in favour of keeping things simple if there's no clear impetus for added complexity.

.taplo.toml
```diff
@@ -3,13 +3,12 @@ include = ["**/*.toml", "**/Cargo.toml", "Cargo.toml"]
 align_comments      = true
 align_entries       = true
 allowed_blank_lines = 1
+indent_entries      = true
```
asmello, line 6: Woah, not sure about this. While any style is technically fine if it's enforced by tooling, this diverges significantly from pretty much every other `Cargo.toml` I've seen... I'd avoid it on a "least surprises" basis.

src/assign.rs
```diff
@@ -1,82 +1,447 @@
...
+
+    #[cfg(test)]
+    mod tests {
+        use core::str::FromStr;
+
+        use serde_json::{json, Value};
+
+        use crate::{Pointer, PointerBuf};
+
+        use super::*;
```
asmello, line 325: nit: spacing

.taplo.toml
chanced: woops, thanks for catching that.

chanced: I just uncommented out things I had tried before, trying to get the alignment right on the object style (which I ended up just backing out anyway)

chanced: fixed, thanks.

src/assign.rs
chanced: Fixed, thanks :)

chanced:
Ugh, so many broken doc tests

chanced:
Walking back the feature flags for this PR.

chanced:
was about to hit Squash and merge  but hit this:

---- pointer::tests::qc_intersection stdout ----
thread 'pointer::tests::qc_intersection' panicked at /Users/chance/.cargo/registry/src/index.crates.io-6f17d22bba15001f/quickcheck-1.0.3/src/tester.rs:165:28:
[quickcheck] TEST FAILED. Arguments: (PointerBuf(""), PointerBuf("/"), PointerBuf("/_\u{10fffd}\u{8d}¡5q^U~1S\u{3e1b3}*5\u{349fe}\u{2029}蔗|\u{d7938}\u{96}[),_峠턲\u{96}\u{858ef}\u{10e26d}帋Ϲ\u{17}R湤E\u{95}7g&\u{10}(\u{98284}㞩\u{e}|\u{8b}¯\u{86}‧;1\u{9fbb2}㓌\u{8}\u{a69da}¤S \u{92}4\u{9efc0}?\u{2068}N_\u{e001}\u{604}\u{10ffff}$?"))

PR #48, opened 2024-06-25
Adds TOML support, refactors tests, adds `Deserialize` to `Pointer`

### LINKED ISSUE #44: `Pointer::intersection` fails if one pointer equals or ends with `"/"`
```rust
    #[test]
    fn test_intersection_issue_44() {
        let base = PointerBuf::new();
        let suffix_0 = Pointer::from_static("/").to_buf();
        let suffix_1 = Pointer::from_static("/a/b/c").to_buf();
        println!("suffix_1: {suffix_1:?}");
        let mut a = base.clone();
        a.append(&suffix_0);
        let mut b = base.clone();
        b.append(&suffix_1);
        let isect = a.intersection(&b);
        assert_eq!(isect, base);
    }
```
**edit:** (above was fixed)

This test is also failing:

```rust
        let base = PointerBuf::must_parse("/a");
        let suffix_0 = PointerBuf::must_parse("/");
        let suffix_1 = PointerBuf::must_parse("/suffix_1");
        let mut a = base.clone();
        a.append(&suffix_0);
        let mut b = base.clone();
        b.append(&suffix_1);
        let isect = a.intersection(&b);
        assert_eq!(
            isect, base,
            "intersection failed\n\n expected: \"{base}\"\n   actual: \"{isect}\""
        );
```
```
 expected: "/a"
   actual: "/a/"
```

chanced:
- Adds TOML support
- Implements `Deserialize` for `Pointer
- Refactors tests for `assign`, `resolve`, `delete`
- Fixes #44


src/pointer.rs
```diff
@@ -99,12 +99,16 @@ const fn is_valid_ptr(value: &str) -> bool {
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
 pub struct Pointer(str);
 
+impl<'a> Default for &'a Pointer {
+    fn default() -> Self {
+        unsafe { &*("" as *const str as *const Self) }
+    }
+}
```
asmello, line 106: I don't get the `'a` lifetime here, this is just `'static`, no?

Also why not use `Pointer::root()` in the body? This is effectively duplicating that code.

src/assign.rs
```diff
@@ -313,26 +313,160 @@ mod json {
...
-                let replaced = ptr.assign(&mut data, value);
+                let replaced = ptr.assign(&mut data, value.clone());
+                assert_eq!(
+                    &expected_data, &data,
+                    "data not as expected.\n\nexpected:\n\n{expected_data}\n\nactual:\n{data}\n"
+                );
+                assert_eq!(
+                    &expected_replaced,
+                    &replaced,
+                    "\nreplaced not as expected for test index {i}:\n\npointer:{ptr}\nvalue:{value}\n\ndata:\n{data}\n\nexpected data:\n{expected_data}\n\nexpected:\n{}\n\nactual:\n{}\n\n",
```
asmello, line 461: I still think `insta` is worth using here. Keeping this up-to-date manually is a pain, and it's hard to read.

src/assign.rs
```diff
@@ -380,7 +514,7 @@ mod json {
             ];
 
             for (path, val, expected, expected_replaced) in tests {
-                let ptr = PointerBuf::parse(path).expect(&format!("failed to parse \"{path}\""));
+                let ptr = PointerBuf::must_parse(path);
```
asmello, line 517: Why a `PointerBuf` here?

src/pointer.rs
```diff
@@ -121,10 +125,21 @@ impl Pointer {
...
         validate(s.as_ref()).map(Self::new)
     }
-
+    /// Attempts to parse a string into a `Pointer`.
+    ///
+    /// If successful, this does not allocate.
+    ///
+    /// ## Panics
+    /// Panics string is not a valid JSON Pointer.
+    pub fn must_parse<S: AsRef<str> + ?Sized>(s: &S) -> &Self {
```
asmello, line 140: This is very similar to `Pointer::from_static`. I'm not opposed to adding it, but at least in the tests I think you can use that instead.

src/pointer.rs
```diff
@@ -595,9 +641,13 @@ impl PointerBuf {
...
+        self._append(other.as_ref())
+    }
+
+    fn _append(&mut self, other: &Pointer) -> &PointerBuf {
         if self.is_root() {
-            self.0.clone_from(&other.0);
+            self.0 = other.0.to_string();
         } else if !other.is_root() {
             self.0.push_str(&other.0);
         }
```
asmello, line 647: Why the split?

src/assign.rs
```diff
@@ -313,26 +313,160 @@ mod json {
...
-            for (mut data, ptr, value) in tests {
+            for (i,Test {
+                mut data,
+                ptr,
+                value,
+                expected_data,
+                expected_result: expected_replaced,
+            }) in tests.into_iter().enumerate()
+            {
                 let ptr = PointerBuf::from_str(ptr).unwrap();
```
asmello, line 452: Missed this on the earlier version, but why `PointerBuf`? It doesn't look like it requires mutation.

src/pointer.rs
chanced: Ugh, not sure why I did that - for whatever reason, I wasn't getting `Pointer::root` autocompletion so I just quickly spit something out.

Thanks for catching it. Changed it to:
```rust
impl Default for &'static Pointer {
    fn default() -> Self {
        Pointer::root()
    }
}
```

src/assign.rs
chanced: I don't know :)

I think the old tests used them because they were ported from the old code. I just kept chugging away with the way things were. Fixing it.

src/pointer.rs
chanced: Oh, because I made `append` generic and I was toying around trying to understand `#[inline(never)]` a bit better. I just forgot to switch it back.

chanced: Thanks for catching it. I'll move it back.

src/pointer.rs
```diff
@@ -315,30 +330,26 @@ impl Pointer {
...
+            } else {
+                break;
             }
-            idx += 1;
-        }
-        if idx == self.0.len() || self.0.as_bytes()[idx] == b'/' {
-            Self::new(&self.0[..idx])
-        } else {
-            Self::new(&self.0[..last_slash])
         }
```
asmello, line 350: I could be wrong, but wouldn't it be simpler to do something like `self.tokens().zip(other.tokens())`?

Also we should definitely add `Token::len`.

src/pointer.rs
```diff
@@ -595,9 +635,13 @@ impl PointerBuf {
...
 
     /// Merges two `Pointer`s by appending `other` onto `self`.
-    pub fn append(&mut self, other: &PointerBuf) -> &PointerBuf {
+    pub fn append<P: AsRef<Pointer>>(&mut self, other: P) -> &PointerBuf {
+        self._append(other.as_ref())
+    }
+
+    fn _append(&mut self, other: &Pointer) -> &PointerBuf {
         if self.is_root() {
-            self.0.clone_from(&other.0);
```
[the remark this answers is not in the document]
chanced: I couldn't get it to work. The previous `other` was a `PointerBuf` which `other.0` would be a `String`. What would the statement look like with `other` being `&Pointer`?

src/assign.rs
chanced: It doesn't. This was another one of those that I think was either a `Pointer` before or I just spaced out and wrote `PointerBuf`.

Thanks for catching it. I got this and the rest.

src/pointer.rs
chanced: Yea, it probably would be easier to use `tokens`.

I keep using `Chars::count` because in most cases, the tokens are looped over using chars... but then I screwed up and used byte index in `split_at` which may be why I'm seeing failing tests that aren't immediately obvious

```rust
    pub fn split_at(&self, idx: usize) -> Option<(&Self, &Self)> {
        if self.0.get(idx..idx + 1) != Some("/") {
            return None;
        }
        let (head, tail) = self.0.split_at(idx);
        Some((Self::new(head), Self::new(tail)))
    }
```

chanced: No, I may be screwing up by using `Chars::count` altogether.

chanced: I'm not sure what the right answer is - to continue using `Chars` or switch to `bytes` everywhere.

chanced: That was it. Switching it to `bytes` flipped it green.

I'll add `len` to `Pointer` and `Token`.

chanced: Actually.. adding `len` to `Token` is going to be confusing, right?

src/pointer.rs
chanced: Ya, it is. I wanted it on `PointerBuf` and copied it over to `Pointer` for uniformity. I'm cool with dumping either or both.

chanced:
I can't figure this one out.

enabling `"toml"` breaks the build because:
```rust
impl PartialOrd<String> for Pointer {
    fn partial_cmp(&self, other: &String) -> Option<Ordering> {
        self.0.partial_cmp(other)
    }
}
```
```
error[E0277]: can't compare `str` with `std::string::String`
   --> src/pointer.rs:539:28
    |
539 |         self.0.partial_cmp(other)
    |                ----------- ^^^^^ no implementation for `str < std::string::String` and `str > std::string::String`
    |                |
    |                required by a bound introduced by this call
    |
    = help: the trait `PartialOrd<std::string::String>` is not implemented for `str`
    = help: the following other types implement trait `PartialOrd<Rhs>`:
              <&'a str as PartialOrd<winnow::stream::BStr>>
              <&'a str as PartialOrd<winnow::stream::Bytes>>
              <str as PartialOrd<winnow::stream::BStr>>
              <str as PartialOrd<winnow::stream::Bytes>>
              <str as PartialOrd>
```

src/assign.rs
chanced: I wasn't sure how obvious it was to tell the expectations of `insta` vs this which you can see the input vs expected output.

chanced:
@asmello I added `toml` support but the only way I could get it to compile is commenting out `impl PartialOrd<String> for Pointer`. I cannot figure out why its causing the build to break.

src/pointer.rs
asmello: I think it's ok to keep both, we should just favour `from_static` because that gives a more convenient lifetime.

src/assign.rs
asmello: The `insta` paradigm is that you never explicitly write the expected output, it will run your test once and record the output in a file, which becomes the ground truth. From that point on, every run of the test will use the previously accepted output, so you get the same effect as using `assert_eq!`, but without cluttering the code with these strings. Then if you ever change the expected output, you can just regenerate the snapshot.

src/pointer.rs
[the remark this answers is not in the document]
asmello: The signature `clone_from(&mut self, source: &Self)`, and `Self` here is `String`, so I'm guessing the issue is you end up with `&str` as `source`, not `&String`. I think it's fine to change it then - we can go from `&String` to `&str` but not the other way around.
 
That said, I wonder if it wouldn't be more idiomatic to follow `Vec::append` and make it:
```
fn append(&mut self, other: &mut PointerBuf)
```

Then make this a flavour of `extend`.

src/pointer.rs
asmello: I think it's fine, it's like how `String::len` returns the number of bytes, not chars, even though not every byte is displayed. We should document it clearly, but the semantics is consistent with `std`.

asmello:
Looking into the `PartialOrd` thing, but I noticed a lot of clippy errors due to `#[deny(clippy::pedantic)]`. Worth fixing them before merging.

chanced:
Awesome, thanks!

Hmm, I wonder why rust analyzer isn't picking up the pedantic lints.

Re: ci, will do. I'm not very experienced with rust ci yet but I needed to update it anyway to handle feature flags and ideally add caching.

asmello:
> Hmm, I wonder why rust analyzer isn't picking up the pedantic lints.

I think you need to enable it with: `"rust-analyzer.checkOnSave.command": "clippy"`

chanced:
Yea, that was it. Thanks!

src/assign.rs
chanced: Right, I did watched the video and read through some of the docs.

I can understand that table-testing is funky to read at first. I absolutely did not care for it when I first saw it. It took exposure and experimentation to finally come around to it.

Setting insta aside, the advantages to it are being able to see the cases covered and expected output at a glance. Ideally helping facilitate tests as docs. They are a pain to setup but ironically allow for pretty fast coverage and edge-case correction. It just takes some getting used to. There is a bit of boilerplate though, so it does clutter things up I guess.

Are they not common in rust? Is insta favored enough that people know to run or review the cases? I'll give it a shot.

Also, sorry about the strings. They were there largely to help me debug as I went through the cases.

asmello: I'm not sure we're on the same page, table-testing and insta aren't either/or at all! But it's my fault, I commented in the wrong place, the hard-to-read string here is not actually the problem, since it's just providing a nicer error message if the test fails, rather than the error message from an `Err` you're expecting. Arguably the solution to make this more readable then is using something like [`indoc`](https://docs.rs/indoc/latest/indoc/) instead. Sorry for the mixup. 

`insta` may still be useful here so that you don't have to explicitly craft `expected_result` and `expected_data`. To be super clear, you'd still have table-testing, that doesn't change at all. It's just how you define the expected output that changes.

I've proposed this because I've found it helps with long term maintenance, requiring less manual intervention if the internal structure of errors change or as new test cases are added, but given how you've defined errors transparently, perhaps it's not worth it. It shines the most when errors are opaque, or you expect complex data that's tricky to craft manually (e.g., a deeply nested JSON). So if you prefer to keep things straightforward, that's totally fine.

chanced: Ahh, okay. I'm game for using `insta`. It's a super cool tool, from the looks of it.

Thank you for taking the time to explain!

src/pointer.rs
chanced: I'm cool with dumping the `must_parse` methods, I don't recall running into too many of them in the wild. Where as `from_static` being `const fn` is awesome and provides a bit more utility than avoiding a call to `unwrap`/`expect`.

chanced:
@asmello can i get you to review the usage of `insta` in `assign` when you have time, please?

Is there a better approach to handling it? I'm not crazy about the formatting of `data` but I can live with it.

chanced:
> I think the assign::toml thing is entirely a red herring, the error doesn't go away if I comment it out. But the fix is simple:

I bet that was because when I was testing it, I only had `assign` done. Once `resolve` and `delete` were written, which I believe both were by the time you had a chance to review, the error remained even if `assign::toml` was commented out.

I tried various approaches to get it to deref but didn't think to use `as_str`. That was a weird one.

asmello:
> I'm not crazy about the formatting of data but I can live with it.

I guess that's because it becomes part of the metadata written to the snapshot header, which is YAML. That's unfortunately not configurable.

As for the body/expression, you can use [a different serialiser](https://insta.rs/docs/serializers/) to get a more pretty-printed output, so there's at least more options.

src/assign.rs
```diff
@@ -300,40 +320,127 @@ mod json {
...
+                tests.into_iter().for_each(Test::run);
+            }
+            fn run(self) {
+                let mut data = self.data.clone();
+
+                let ptr = Pointer::from_static(self.ptr);
+                let replaced = ptr.assign(&mut data, self.assign.clone());
+                let mut s = insta::Settings::new();
+
+                s.set_info(&self);
```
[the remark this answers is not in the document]
chanced: Yea, I caught that as I was rewriting. I was trying to figure out why I couldn't set the `info`. 

Turned out I needed `"serde"` feature flag.

chanced:
@asmello i had to remove the tables for it to work. Code has been updated to reflect it.

chanced:
Well.. I thought it was fixed. CI is showing failures. Hmmm..

asmello:
> @asmello i had to remove the tables for it to work. Tests in assign have been updated.

That's... unexpected. It should work fine with the table tests. I've done it like that before.

> Well.. I thought it was fixed. CI is showing failures. Hmmm..

I think the snapshots are just out of date, you need to regenerate them.

chanced:
It's still hitting weird inconsistencies, for example:
Screenshot 2024-06-27 at 2 41 51 PM

The discrepancy between new and old is weird. The old snapshot is the result from comparing the replaced result while the new is comparing data.

I've run it a few times. I'll delete the snapshots and try again.

asmello:
Oh interesting, I wonder if it's having trouble assigning the outputs to a number deterministically. I think you were calling the tests in a loop right? I recall there was a fix for that.

chanced:
I was, which I assumed was the problem. I've since changed it so that they aren't called from a loop. However, the calls to `assert_snapshot(data)` and `assert_debug_snapshot(result)` are happening in a `fn` now:

```rust
        #[derive(Serialize, Debug)]
        struct Test {
            data: Value,
            ptr: &'static str,
            assign: Value,
        }

        fn test(data: impl Into<Value>, ptr: &'static str, assign: impl Into<Value>) {
            let mut data = data.into();
            let assign = assign.into();

            let test = Test {
                ptr,
                data: data.clone(),
                assign: assign.clone(),
            };
            let ptr = Pointer::from_static(test.ptr);
            let replaced = ptr.assign(&mut data, assign);
            insta::with_settings!({
                info => &test,
                description => "asserting that the JSON data is as expected after assignment",
                omit_expression => true // do not include the default expression
            }, {
                assert_snapshot!(&data);
            });
            insta::with_settings!({
                info => &test,
                description => "asserting that the replaced JSON data returned from the assignment is as expected",
                omit_expression => true // do not include the default expression
            }, {
                assert_debug_snapshot!(&replaced);
            });
        }
```

where tests are executed serially:

```rust
            test(json!({}), "/foo", json!("bar"));
            test(json!({"foo": "bar"}), "", json!("baz"));
            test(json!({"foo": "bar"}), "/foo", json!("baz"));
            test(json!({"foo": "bar"}), "/foo/bar", json!("baz"));
```

asmello:
Yeah, it's weird that still fails. Maybe it's clashing with the old outputs somehow? I'd expect deleting and starting over to fix in that case.

The way to make it work with loops is similar to [their recipe for rtest](https://insta.rs/docs/patterns/): it's using [`set_snapshot_suffix`](https://docs.rs/insta/latest/insta/struct.Settings.html#method.set_snapshot_suffix).

chanced:
Using simple indexes didn't fly.

It's definitely because they are being called from a fn rather than inline. I guess it'd need a unique name for the test set + an index. The `"Snapshot File"` suggests to me they are colliding:

Screenshot 2024-06-27 at 3 13 18 PM

chanced:
> BTW, I'm sorry this is causing you trouble. 😓 I've never had a hard time with insta... Do feel free to drop it if you get frustrated.

Oh, it's no worries! New tools in this industry often come with a bit of frustration :). Also, this is very much seems to be caused by me not using it properly.

Here is the updated `test` `fn`:

```rust
        fn test(idx: usize, data: impl Into<Value>, ptr: &'static str, assign: impl Into<Value>) {
            let mut data = data.into();
            let assign = assign.into();

            let test = Test {
                ptr,
                data: data.clone(),
                assign: assign.clone(),
            };
            let ptr = Pointer::from_static(test.ptr);
            let replaced = ptr.assign(&mut data, assign);
            insta::with_settings!({
                snapshot_suffix => format!("{idx}-data"),
                info => &test,
                description => format!("case #{idx}: asserting that the JSON data is as expected after assignment"),
                omit_expression => true // do not include the default expression
            }, {
                assert_snapshot!(&data);
            });
            insta::with_settings!({
                info => &test,
                snapshot_suffix => format!("{idx}-result"),
                description => format!("case #{idx}: asserting that the replaced JSON data returned from the assignment is as expected"),
                omit_expression => true // do not include the default expression
            }, {
                assert_debug_snapshot!(&replaced);
            });
        }
```

chanced:
@asmello I was able to get it to work with a `macro_rules` but its just not really my cup of tea.

I removed insta, rewrote a number of the tests to conform to a single approach, and refactored the mod structure of `assign`, `delete`, `resolve` a bit. 

Hopefully this makes it a bit less cluttered and distracting.

asmello:
That's alright, thanks for giving it a try and I'm sorry it didn't pan out.

I'll give it another look, but feel free to merge when you feel ready.

src/pointer.rs
```diff
@@ -1,84 +1,27 @@
...
+#[cfg(feature = "resolve")]
 use crate::resolve::{Resolve, ResolveMut};
 use crate::{InvalidEncodingError, ParseError, ReplaceTokenError, Token, Tokens};
 use alloc::{
     borrow::ToOwned,
     string::{String, ToString},
     vec::Vec,
 };
+use core::fmt;
 use core::{borrow::Borrow, cmp::Ordering, ops::Deref, slice, str::FromStr};
```
asmello, line 14: Tiny nit: these can be merged.

src/pointer.rs
```diff
@@ -99,32 +42,50 @@ const fn is_valid_ptr(value: &str) -> bool {
...
     }
 }
-
 impl Pointer {
     /// Private constructor for strings that are known to be correctly encoded
     ///
     /// This is a zero-copy constructor
     fn new<S: AsRef<str> + ?Sized>(s: &S) -> &Self {
-        unsafe { &*(s.as_ref() as *const str as *const Self) }
+        unsafe { &*(core::ptr::from_ref::<str>(s.as_ref()) as *const Self) }
```
asmello, line 60: Out of interest, was this clippy's suggestion? I originally copied the expression from `std::path::Path` essentially verbatim, so I'm amused clippy wasn't happy with it :p

src/index.rs
```diff
@@ -122,6 +127,7 @@ impl Index {
     /// assert_eq!(Index::Num(42).for_len_unchecked(30), 42);
     /// assert_eq!(Index::Next.for_len_unchecked(30), 30);
     /// ````
+    #[must_use]
```
asmello, line 130: Any particular reason to make this one `must_use` and not the others?

src/pointer.rs
```diff
@@ -99,32 +42,50 @@ const fn is_valid_ptr(value: &str) -> bool {
...
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         self.0.fmt(f)
     }
 }
-
 impl Pointer {
     /// Private constructor for strings that are known to be correctly encoded
     ///
     /// This is a zero-copy constructor
     fn new<S: AsRef<str> + ?Sized>(s: &S) -> &Self {
```
asmello, line 59: I wonder if this and the other constructors should be `must_use` too?

`must_parse` probably can be left out, as I can imagine using it as a validator being useful in tests.

src/pointer.rs
```diff
@@ -315,39 +324,33 @@ impl Pointer {
...
-            } else {
-                return self;
-            }
+        if self.is_root() || other.is_root() {
+            return Self::root();
         }
-        let mut last_slash = 0;
         let mut idx = 0;
-        for (a, b) in self.0.bytes().zip(other.0.bytes()) {
+        for (a, b) in self.tokens().zip(other.tokens()) {
```
asmello, line 333: I like how simple and neat this turned out!

src/pointer.rs
```diff
@@ -738,16 +828,114 @@ impl core::fmt::Display for PointerBuf {
...
     use super::*;
     use quickcheck::TestResult;
     use quickcheck_macros::quickcheck;
 
     #[test]
-    #[should_panic]
+    #[should_panic = "invalid JSON Pointer"]
     fn from_const_validates() {
-        Pointer::from_static("foo/bar");
+        let _ = Pointer::from_static("foo/bar");
```
asmello, line 938: I know I told you we should prefer `from_static`, but `must_parse` reads a bit better when used as a validator... And it makes sense to not make it `must_use`. Consider this a nit though.

asmello (commented):
essentially nits, looking great! awesome job tidying things up

asmello (approved):

src/pointer.rs
chanced: It was a clippy lint, ya. I wasn't sure about if I should apply the suggestion. I know very little working in unsafe. I probably should have avoided it. 

Sorry about that.

src/index.rs
chanced: If the others are `Option` or `Result`, then they have it at the type level. 

I'll look closer when I'm at my computer.

src/pointer.rs
chanced: Yea, I was surprised how clean it came out. I definitely borrowed from your initial implementation. 😀

src/pointer.rs
asmello: No, no, no. This is perfectly fine, I was just curious. I'd expect std to apply clippy lints themselves, but maybe it's a new lint or something.

src/index.rs
asmello: Ah right, that's probably what it is. 👍

src/pointer.rs
asmello: Also, I should probably have left a link to the reference implementation: https://doc.rust-lang.org/src/std/path.rs.html#1-3385

I'm no unsafe expert either, but I think this one is straightforward. It's just transmuting from a reference to an `str` to a reference to a `Pointer`, without copying, which relies on the internal representation of either being the same. Writing this down just now made me realise I missed one important detail in the definition of `Pointer`: `#[repr(transparent)]`. Will push a small PR to fix that.

src/pointer.rs
chanced: Hah, I was just about to rip out `must_parse`.

Hmm.

src/pointer.rs
asmello: https://github.com/chanced/jsonptr/pull/49

src/pointer.rs
asmello: Well arguably we could also just use `Pointer::parse(...).unwrap()`, which is also pretty idiomatic if a little more verbose. Up to you.

src/pointer.rs
chanced: Sorry, I was wrapping up some errands earlier.

This is the lint that was tripping: https://rust-lang.github.io/rust-clippy/master/index.html#ref_as_ptr

> # ref_as_pt
> What it does
>
> Checks for casts of references to pointer using as and suggests `std::ptr::from_ref` and `std::ptr::from_mut` instead.
> 
> Why is this bad?
> Using as casts may result in silently changing mutability or type.
>
> Example
> ```rust
> let a_ref = &1;
> let a_ptr = a_ref as *const _;
> ```
> Use instead:
> ```rust
> let a_ref = &1;
> let a_ptr = std::ptr::from_ref(a_ref);
>

src/index.rs
chanced: Yep, this was it. Great eye tho! I doubt I would ever have noticed that :)

src/pointer.rs
asmello: Added in 1.78.0. Yep, pretty new. Thanks.

src/pointer.rs
chanced: I took out `must_parse`, left `from_static`.

src/pointer.rs
chanced: Ah, I forgot about this one.

So after reading the lint description again, I'm actually okay with turning it off and rolling back all of the `#[must_use]` tags. They were added due to the pedantic [`clippy::must_use`)(https://rust-lang.github.io/rust-clippy/master/#/must_use) lint.

PR #49, opened 2024-06-29
add repr transparent to `Pointer`

asmello:
I missed this when first implementing the slice-like `Pointer` type. I _think_ the compiler will use the same layout as `str` regardless, but it's important to make sure. This assumption is needed for the unsafe code used in this type to be sound.


chanced:
Awesome!

chanced:
Awesome! Should we need to do the same for `PointerBuf`?

asmello:
No, `PointerBuf` doesn't have any unsafe code that relies on the internal layout of the type. 👍

chanced:
Awesome, thanks. 

heh, as an aside, I really need to get better about proofreading my comments written on my phone. So many incoherent sentences 🤦‍♂️

PR #54, opened 2024-07-02
Fix readme shields

chanced:
- fixes build shield, it now runs against test.yml
- adds codecov shield


chanced:
oops, branched from the wrong branch.

PR #56, opened 2024-07-02
Improve docs & codecov for 0.5.0

chanced:
- improves documentation
- brings code coverage to 99%
- fixes `TryFrom` for `Index`
- adds `FromStr` impl for `Index`


chanced:
@asmello I think flattening the crate, except for `resolve`, `assign`, `delete`, where the traits are re-exported at the root. This means eliminating the `index` though, so we would need to move the excellent documentation you've already written.

The reasoning for this is that I've found folks tend to prefer well structured hierarchy when working on project but prefer a flattened API when utilizing a library. I'm still rather novice to rust, so it could be that Rustaceans disagree with this. If you feel like that's the case, I think having a `token` mod that houses `Token`, `Tokens`, `Index`, and any respective errors. I don't know where we'd put `Component` and `Components` though.

I had gotten the code coverage to 99% before introducing `Component` / `Components`. Both are untested (and undocumented) so we are sitting at 98.2%. I'll get it back up over 99% before this release is cut. I don't know if I can hit 100% tho.

I've written some crate-level docs. They are currently being imported from the `README.md` but I plan on changing that. I'm just using markdown for ease of writing, formatting, and highlighting. Once the verbiage is settled, I'll move them to doc comments and make the `README.md` more generic.

I'm headed out to the river but I should hopefully be able to get this finished over the weekend or early next week.

asmello:
> @asmello I think flattening the crate, except for `resolve`, `assign`, `delete`, where the traits are re-exported at the root, makes sense. This means eliminating the `index` though, so we would need to move the excellent documentation you've already written.

Not sure I follow what you're saying. I'm ok with having the commonly used symbols exported at the root; that is common practice in library crates. If not best practice. This doesn't preclude having a hierarchical project structure though, and that is also best practice. Even if a module is not public it makes sense for it to be relatively well documented.

One rule I'd hold though is to only _publicly_ export each symbol at one path. This can be the root, but if it is then module the symbol comes from shouldn't also be public.

> I had gotten the code coverage to 99% before introducing Component / Components. Both are untested (and undocumented) so we are sitting at 98.2%. I'll get it back up over 99% before this release is cut. I don't know if I can hit 100% tho.

I think not hitting 100% is ok. Lack of coverage indicates code isn't tested, which isn't good, but having coverage doesn't prove tests are useful, only that they exist. So while having a relatively high coverage is a worthy objective, it gives decreasing returns past a certain point.

> Once the verbiage is settled, I'll move them to doc comments and make the README.md more generic.

I actually like this pattern of having doc-tests in the README. It encourages keeping the README up-to-date as code changes, and lots of crates do this. Is there a particular reason you want to avoid it?

BTW minor point of feedback, but I'd avoid chore-style PRs like this from also including new code like `Components`. I can handle the diff, but it makes tracking down bugs harder down the line. I also find small PRs useful for contributors to learn from.

chanced:
> Not sure I follow what you're saying. I'm ok with having the commonly used symbols exported at the root; that is common practice in library crates. If not best practice. This doesn't preclude having a hierarchical project structure though, and that is also best practice. Even if a module is not public it makes sense for it to be relatively well documented.

Oh, I didn't mean to flatten it all into `lib.rs`. I considered it initially, because I've read some people prefer that, but I have an increasingly hard time navigating massive documents. I just meant publicly exposed. So the only 3 mods would be `assign`, `resolve`, and `delete`. I understand that my understanding of hierarchy preferences may not match up in rust.

> BTW minor point of feedback, but I'd avoid chore-style PRs like this from also including new code like Components. I can handle the diff, but it makes tracking down bugs harder down the line. I also find small PRs useful for contributors to learn from.

Yea, I regret this. The only reason I did was because I wanted to be able to reference `Component`/`Components` in the docs and they didn't exist. I'll delete them from this PR and introduce them.

> I actually like this pattern of having doc-tests in the README. It encourages keeping the README up-to-date as code changes, and lots of crates do this. Is there a particular reason you want to avoid it?

The reason I was considering changing this was that the README.md could be made more generic, using terms such as "for Rust," without seeming out of place. All of my projects up to this point have used README and I genuinely love the practice - especially that it keeps the README in check.

Scratch the plan to move the docs out of README.md. We will keep that practice.

asmello:
> Oh, I didn't mean to flatten it all into `lib.rs`. I considered it initially, because I've read some people prefer that, but I have an increasingly hard time navigating massive documents. I just meant publicly exposed. So the only 3 mods would be `assign`, `resolve`, and `delete`. I understand that my understanding of hierarchy preferences may not match up in rust.

If you mean only those two would be exported as wholesale mods, that's probably ok. I don't think there are enough types that we _need_ a hierarchy in the public API.

> Yea, I regret this. The only reason I did was because I wanted to be able to reference `Component`/`Components` in the docs and they didn't exist. I'll delete them from this PR and introduce them.

No need to revert now, just for future reference.

> The reason I was considering changing this was that the README.md could be made more generic, using terms such as "for Rust," without seeming out of place. All of my projects up to this point have used README and I genuinely love the practice - especially that it keeps the README in check.
> 
> Scratch the plan to move the docs out of README.md. We will keep that practice.

Yeah it's a bit of a mixed bag, I've also seen projects use a fancy README that doesn't really work well in docs.rs... I guess we don't need to use the same README for docs.rs and for Github, let me think a bit.

asmello:
I think we can take a clue from here: https://linebender.org/blog/doc-include/

They have [a single README.md](https://github.com/linebender/android_trace/blob/f8b46a95311a6d5ce0409fffdc77f5919084f170/android_trace/README.md#L102) that renders differently in [docs.rs](https://docs.rs/android_trace/0.1.1/android_trace/) and in [Github](https://github.com/linebender/android_trace/blob/v0.1.1/android_trace/README.md). The main trick seems to be using [some inline html tags with `<div class="rustdoc-hidden">`](https://github.com/linebender/android_trace/blob/v0.1.1/android_trace/README.md?plain=1#L102). Seems a bit more advanced so maybe leave it for later.

chanced:
Oh that's cool! I agree though, let's just have the README pull double duty.

chanced:
> Lack of coverage indicates code isn't tested, which isn't good, but having coverage doesn't prove tests are useful, only that they exist. So while having a relatively high coverage is a worthy objective, it gives decreasing returns past a certain point.

Absolutely. I don't believe that having a really high level of code coverage offers the sort of guarantees on correctness that  it may seem. I do tend to shake out small issues when I strive to hit above > 90% though.

chanced:
@asmello this is ready for you to review whenever you have time.

chanced:
Hmm, looking at the README, I'm kinda second guessing not using the strategy you linked. 

Screenshot 2024-07-08 at 12 46 02 PM

chanced:
I went ahead and employed the rustdoc hack.

src/token.rs
```diff
@@ -315,6 +318,72 @@ impl alloc::fmt::Display for Token<'_> {
...
+╔══════════════════════════════════════════════════════════════════════════════╗
+║                                                                              ║
+║                             InvalidEncodingError                             ║
+║                            ¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯                            ║
+╚══════════════════════════════════════════════════════════════════════════════╝
+░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░
+*/
+
+/// A token within a json pointer contained invalid encoding (`~` not followed
+/// by `0` or `1`).
```
asmello, line 360: Worth adding a comment somewhere (probably not a doc-comment though) about why this is the only case that can occur. In particular, why we never produce a `/`, which would also be invalid inside a token.

src/resolve.rs
```diff
@@ -4,9 +4,27 @@
...
-//! ## Feature Flag
 //! This module is enabled by default with the `"resolve"` feature flag.
 //!
+//! ## Usage
+//! [`Resolve`] and [`ResolveMut`] can be used directly or through the
+//! [`resolve`](Pointer::resolve) and [`resolve_mut`](Pointer::resolve_mut)
+//! methods on [`Pointer`] and [`PointerBuf`](crate::PointerBuf).
+//! ```rust
+//! use jsonptr::{Pointer, Resolve, ResolveMut};
+//! use serde_json::json;
```
asmello, line 15:
```suggestion
//! # use jsonptr::{Pointer, Resolve, ResolveMut};
//! # use serde_json::json;
```
?

src/resolve.rs
```diff
@@ -4,9 +4,27 @@
...
+//!
+//! let ptr = Pointer::from_static("/foo/1");
+//! let mut data = json!({"foo": ["bar", "baz"]});
+//!
+//! let value = ptr.resolve(&data).unwrap();
+//! assert_eq!(value, &json!("baz"));
+//!
+//! let value = data.resolve_mut(ptr).unwrap();
+//! assert_eq!(value, &json!("baz"));
+//!
```
asmello, line 25:
```suggestion
//! assert_eq!(value, &json!("baz"));
```
:p

src/resolve.rs
```diff
@@ -4,9 +4,27 @@
...
 //! Pointer.
 //!
-//! ## Feature Flag
 //! This module is enabled by default with the `"resolve"` feature flag.
 //!
+//! ## Usage
+//! [`Resolve`] and [`ResolveMut`] can be used directly or through the
+//! [`resolve`](Pointer::resolve) and [`resolve_mut`](Pointer::resolve_mut)
+//! methods on [`Pointer`] and [`PointerBuf`](crate::PointerBuf).
+//! ```rust
```
asmello, line 13:
```suggestion
//! methods on [`Pointer`] and [`PointerBuf`](crate::PointerBuf).
//!
//! ```rust
```

src/resolve.rs
```diff
@@ -395,55 +416,198 @@ mod toml {
...
-        }
+    #[test]
+    fn resolve_error_display() {
+        let err = ResolveError::FailedToParseIndex {
+            offset: 0,
+            source: ParseIndexError {
+                source: "invalid".parse::<usize>().unwrap_err(),
+            },
+        };
+        assert_eq!(format!("{err}"), "failed to parse index at offset 0");
```
asmello, line 461: Isn't this just `err.to_string()`?

src/resolve.rs
```diff
@@ -395,55 +416,198 @@ mod toml {
...
+    };
     use core::fmt;
 
-    struct Test<'v, V> {
-        ptr: &'static str,
-        expected_result: Result<&'v V, ResolveError>,
-        data: &'v V,
+    #[cfg(feature = "std")]
+    #[test]
+    fn resolve_error_source() {
```
asmello, line 427: Not sure I get the point of this test. Not a problem per se, but feels like it'll just cause maintenance burden down the line.

src/resolve.rs
```diff
@@ -395,55 +416,198 @@ mod toml {
...
 
-            // testing ResolveMut
-            let res = data.resolve_mut(ptr).cloned();
-            assert_eq!(
-                &res, &expected_result,
-                "test #{i} failed:\n\nexpected\n{expected_result:#?}\n\nactual:\n{res:#?}",
-            );
-        }
+    #[test]
+    fn resolve_error_offset() {
```
asmello, line 484: I think this one should use a different value to be more robust, 0 is more likely to be returned "by accident" than most values.

Also this is the sort of thing that would be more useful to check in a more end-to-end test, but no need to change it now.

src/resolve.rs
```diff
@@ -476,86 +640,86 @@ mod tests {
             " ": 7,
             "m~n": 8
         });
-        // let data = &test_data;
+        // let data = &data;
```
asmello, line 643: Is this left on purpose? (here and elsewhere)

src/lib.rs
asmello, line 1: Awesome, this feels cleaner.

src/pointer.rs
```diff
@@ -385,6 +401,24 @@ impl Pointer {
     {
         dest.assign(self, src)
     }
+
+    /// Returns [`Components`] of this JSON Pointer.
+    ///
+    /// A [`Component`](crate::Component) is either [`Token`] or the root
+    /// location of a document.
+    /// ## Example
```
asmello, line 409:
```suggestion
    /// ## Examples
```

src/pointer.rs
```diff
@@ -385,6 +401,24 @@ impl Pointer {
...
+    /// ```
+    /// # use jsonptr::{Component, Pointer};
+    /// let ptr = Pointer::parse("/a/b").unwrap();
```
asmello, line 412:
```suggestion
    /// let ptr = Pointer::from_static("/a/b");
```
?

src/pointer.rs
```diff
@@ -441,13 +475,29 @@ impl PartialEq<&str> for Pointer {
...
+}
 impl PartialEq<str> for Pointer {
     fn eq(&self, other: &str) -> bool {
         &self.0 == other
     }
 }
 
+impl PartialEq<Pointer> for &str {
+    fn eq(&self, other: &Pointer) -> bool {
+        *self == (&other.0)
```
asmello, line 491:
```suggestion
        *self == &other.0
```
or are the parentheses necessary?

src/pointer.rs
```diff
@@ -666,7 +802,7 @@ impl PointerBuf {
             });
         }
         let mut tokens = self.tokens().collect::<Vec<_>>();
-        if index > tokens.len() {
+        if index >= tokens.len() {
```
asmello, line 805: Good find! Is this caught by one of the new tests?

src/pointer.rs
```diff
@@ -845,6 +981,153 @@ const fn validate(value: &str) -> Result<&str, ParseError> {
...
+                )
+            }
+            Self::InvalidEncoding { source, .. } => write!(f, "{source}"),
+        }
+    }
+}
+
+impl ParseError {
+    /// Returns `true` if this error is `NoLeadingBackslash`; otherwise returns
+    /// `false`.
```
asmello, line 1027: You need this to be a single line for rustdoc to parse it properly. Otherwise it'll render weirdly. Ditto elsewhere (picked this location semi-randomly).

src/pointer.rs
```diff
@@ -845,6 +981,153 @@ const fn validate(value: &str) -> Result<&str, ParseError> {
...
+    /// Offset of the first invalid encoding from within the pointer.
+    /// ```text
+    /// "/foo/invalid~tilde/invalid"
+    ///              ↑
+    ///             12
+    /// ```
+    /// ```
+    /// # use jsonptr::PointerBuf;
+    /// let err = PointerBuf::parse("/foo/invalid~tilde/invalid").unwrap_err();
+    /// assert_eq!(err.pointer_offset(), 4)
```
asmello, line 1085: Looks like this example is mismatched

src/pointer.rs
```diff
@@ -906,18 +1222,54 @@ mod tests {
...
+        assert_eq!(err.pointer_offset(), 4);
+        assert_eq!(err.source_offset(), 8);
+        assert_eq!(err.complete_offset(), 12);
+
+        let err = Pointer::parse("invalid~encoding").unwrap_err();
+        assert_eq!(err.pointer_offset(), 0);
+        assert_eq!(err.source_offset(), 0);
+
+        let err = Pointer::parse("no-leading/slash").unwrap_err();
+        assert!(err.source().is_none());
```
asmello, line 1241: Not sure if this should be tested, is adding a source a breaking change? Ditto elsewhere.

src/pointer.rs
```diff
@@ -906,18 +1222,54 @@ mod tests {
...
+        assert!(err.source().is_some());
+        let source = err.source().unwrap();
+        assert!(source.is::<InvalidEncodingError>());
+
+        let err = Pointer::parse("no-leading/slash").unwrap_err();
+        assert!(err.source().is_none());
+    }
+
+    #[test]
+    fn as_pointer() {
```
asmello, line 1258: This name doesn't match the test very well

src/pointer.rs
```diff
@@ -1130,6 +1597,45 @@ mod tests {
         }
     }
 
+    #[test]
+    #[allow(clippy::useless_asref)]
```
asmello, line 1601: Worth a short comment as to why this is a false positive

asmello (commented):
Awesome docs! I'm a bit sleepy but I don't think I would have missed anything major.

I feel like a few tests are too trivial, which is what I feared was going to happen when trying to reach near 100% coverage. But fine, leave them, it's good marketing I guess. :p

README.md
```diff
@@ -1,131 +1,160 @@
...
-Data structures and logic for resolving, assigning, and deleting by JSON Pointers ([RFC
-6901](https://datatracker.ietf.org/doc/html/rfc6901)).
-
-## Usage
-
-JSON Pointers can be created either with a slice of strings or directly from a properly encoded string representing a JSON Pointer.
+JSON Pointers ([RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901))
+defines a string syntax for identifying a specific location within a JSON, or
+similar, document. This crate provides two types, [`Pointer`] and [`PointerBuf`]
+(akin to [`str`] and [`String`]), for working with them abstractly.
```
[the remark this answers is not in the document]
chanced: Hah, I actually started with `Path` and `PathBuf`!

I wafered on it and went with `str`/`String` for familiarity sake. I'll switch it back. I'm with you,  personally prefer `Path` and `PathBuf`.

README.md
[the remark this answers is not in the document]
chanced: Thanks. I'm not too crazy about all the links in the readme tho. It's a bit noisy:

Screenshot 2024-07-08 at 4 00 06 PM

src/delete.rs
```diff
@@ -183,7 +183,7 @@ mod tests {
         fn all(tests: impl IntoIterator<Item = Test<V>>) {
             tests.into_iter().enumerate().for_each(|(i, t)| t.run(i));
         }
-        fn run(self, i: usize) {
+        fn run(self, _i: usize) {
```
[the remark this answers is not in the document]
chanced: So initially the `i` was for the index of the test itself. It was primarily added so that I could easily reconcile which test was failing. I left it in the event we run into failing tests in the future.

I guess the need for `i` kind of demonstrates the short-comings of table-based tests. Had they all been independent tests, this would have been a non-issue.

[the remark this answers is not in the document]
chanced: I forgot to mention why it became `_i`. The usage of `i` was used in the output of the `assert` calls. They were pulling down the code coverage and I figured if they are all passing, no need for the message... sigh, my silly ass attempt to reach 100.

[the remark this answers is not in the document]
asmello: The `i` is fine, although perhaps having descriptive labels would be best. Not something to change in this PR, but worth thinking about if you make it a trait in the future.

src/pointer.rs
chanced: Really? Is there a flag or something that I need to enable for it to respect markdown newlines on docs.rs? I thought I had them all enabled locally. If not, I think it respects it:
Screenshot 2024-07-08 at 4 07 09 PM

src/pointer.rs
chanced: Wow, great catch. Damn you're great at this.

chanced:
> I feel like a few tests are too trivial, which is what I feared was going to happen when trying to reach near 100% coverage. 

Yea, my silly ass pursuit of 100 definitely resulted in some obnoxious tests. There were a few tests which I genuinely did not want to write. There were a few, mostly around error outputs, that I let copilot spit out. Yet, I pushed ahead thinking I _might_ actually be able to hit 100 on this lib.

As annoying and useless as some of the tests are, one of the tests that I absolutely felt was useless caught an issue with `TryFrom`.

I'm okay with dumping the useless ones. You've marked a few of them.

src/pointer.rs
asmello: Hmmm so when I ask `rust-analyzer` for inlay docs it shows me a pop-up with a hard break at the newline. Doesn't seem to render the second line differently, so not sure how much of a concern this is really, but Jon once said this could cause issues in places as the first line is indexed specially.

src/token.rs
chanced: Ah, great point.

Hmm, should `token` be `pub mod`?

src/pointer.rs
chanced: Oh that's unfortunate. I use a plugin to linebreak my comments and I guess I shouldn't do that anymore. I had no idea that any tooling in the ecosystem treated newlines as hard breaks. I've always assumed it respected markdown rules.

btw, I'm reluctant to respond knowing you're tired. Go get some rest. This can wait.

chanced: Aw man, I have so many comments to fix. I'm more bummed that I can't auto-format my doc comments anymore. 

Strangely enough, vscode gets it right. 
Screenshot 2024-07-08 at 4 35 24 PM

Btw, is that neovim? 

> Jon also said is that the first line "shows up in more places", though I never figured out which places those are, exactly. 😅

lol!

chanced: I think this one's okay. It seems to be a bug in RA not properly rendering newlines. Which is weird, you'd think that'd have been caught by now.

I wonder what else could possibly be incorrectly parsing markdown and only respecting the first actual line rather than the first paragraph.

src/index.rs
```diff
@@ -184,11 +200,220 @@ macro_rules! derive_try_from {
...
+
+    #[test]
+    fn index_for_len_incl_valid() {
+        assert_eq!(Index::Num(0).for_len_incl(1), Ok(0));
+        assert_eq!(Index::Next.for_len_incl(2), Ok(2));
+    }
+
+    #[test]
+    fn index_for_len_incl_out_of_bounds() {
+        assert!(Index::Num(2).for_len_incl(1).is_err());
```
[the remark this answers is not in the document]
chanced: Ohhh, I forget that tests fail if there's a panic. I like that.

src/pointer.rs
chanced: Agreed, renamed to `pointerbuf_as_pointer_returns_pointer`

src/pointer.rs
chanced: 👍

src/pointer.rs
chanced: It was to get coverage on that branch of `source` 😅. I don't know if it'd be considered a breaking change but I would assume so?

src/pointer.rs
chanced: I wish there was a way to turn an entire discussion into an issue. I've update some of them, but not all.

src/delete.rs
```diff
@@ -212,7 +204,7 @@ mod tests {
     */
     #[test]
     #[cfg(feature = "json")]
-    fn test_delete_json() {
+    fn delete_json() {
```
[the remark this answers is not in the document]
chanced: heh, `sed -i "" 's/test_//g' **/*.rs`

src/pointer.rs
chanced: https://github.com/chanced/jsonptr/issues/58

src/pointer.rs
chanced: Created an issue for this one too: https://github.com/chanced/jsonptr/issues/57

chanced:
I created two issues for the last outstanding conversations: https://github.com/chanced/jsonptr/issues/57 and https://github.com/chanced/jsonptr/issues/58

src/resolve.rs
chanced: thanks!

src/resolve.rs
chanced: yea, good call.

PR #61, opened 2024-07-09
Small DX changes + more docs

chanced:
- renames `PointerBuf::replace_token` to `replace`, `ReplaceTokenError` to `ReplaceError`
- changes `PointerBuf::push_back`, `PointerBuf::push_front`, `PointerBuf::replace` to accept `T: Into<Token>`
- implements `From` all integer types for `Token`
- cleans up and adds docs

### DIFF
--- a/LICENSE-APACHE
+++ b/LICENSE-APACHE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+1. Definitions.
+
+    "License" shall mean the terms and conditions for use, reproduction,
+    and distribution as defined by Sections 1 through 9 of this document.
+
+    "Licensor" shall mean the copyright owner or entity authorized by
+    the copyright owner that is granting the License.
+
+    "Legal Entity" shall mean the union of the acting entity and all
+    other entities that control, are controlled by, or are under common
+    control with that entity. For the purposes of this definition,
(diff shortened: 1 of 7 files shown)


chanced:
@asmello any objection to renaming `replace_token` to `replace`, `ReplaceTokenError` to `ReplaceError`? All other methods omit `token` so I figured it'd align a bit better. The only hesitation I have is that `replace` for string types may give a slightly different expectation.

Also, are you opposed to implementing `PartialEq<&str>` for `Token` (using the decoded value)?

chanced:
This PR will be the second to last I want to finish.

The next will contain a small change to `Component`, adding `offset` as well as adding `walk_to` / `walk_form` to `Resolve` with provided implementations.

chanced:
oh, this was meant to be in draft mode.

asmello:
> Also, do you have any opposition to me implementing PartialEq<&str> for Token (using the decoded value)?

I'm a bit on the fence on this one. While the improved ergonomics is nice, I'm a little afraid people won't realise that it uses the encoded version until they run into an unexpected edge case at runtime. Seems a bit error prone. By forcing straight `Token` and `Token` comparison there's no room for error.

I can change my mind on this, if it turns out to be a major pain point, but I'd err on the side of safety at first.

chanced:
Sounds good to me.

asmello:
Also worth adding a comment so we remember we deliberately chose not to implement this.

chanced:
@asmello i went ahead and removed serde from `Token`.

asmello:
> @asmello i went ahead and removed serde from `Token`.

Cool, I think that's reasonable. It's not immediately useful and there's ambiguity with encoding. 👍

PR #62, opened 2024-07-10
modifies components + adds walk

chanced:
So I'm not sure about this pull request at all. The original idea was to add `walk_to` and `walk_from` to `Resolve` / `ResolveMut`. I'm not crazy about the signature: 

```rust
    fn walk_to<F, B>(&self, ptr: &Pointer, f: F) -> Result<ControlFlow<B>, Self::Error>
    where
        F: Fn(Component, &Self::Value) -> ControlFlow<B>,
    {
        let _ = self.resolve(ptr)?;
        let root = self.resolve(Pointer::root()).unwrap();
        let mut last = None;
        for component in ptr.components() {
            match component {
                Component::Root => {
                    last = Some(f(Component::Root, root));
                }
                Component::Token {
                    token,
                    index,
                    offset,
                    pointer,
                } => {
                    let value = self.resolve(pointer).unwrap();
                    let tok_len = token.encoded().len();
                    let ctrl_flow = f(
                        Component::Token {
                            token,
                            pointer,
                            index,
                            offset,
                        },
                        value,
                    );
                    if offset + tok_len >= ptr.len() {
                        // last token
                        return Ok(ctrl_flow);
                    }
                    if ctrl_flow.is_break() {
                        return Ok(ctrl_flow);
                    }
                    last = Some(ctrl_flow);
                }
            }
        }
        Ok(last.unwrap())
    }
```

I'm tempted to rip out `Components` / `Component`. Initially I thought this would cut down on a good bit of code in my json schema compiler but I'm doubting whether its worth it. 

Would love your feedback when you get time @asmello

### DIFF
--- a/src/component.rs
+++ b/src/component.rs
@@ -2,40 +2,59 @@ use crate::{Pointer, Token, Tokens};
 
 /// A single [`Token`] or the root of a JSON Pointer
 #[derive(Debug, PartialEq, Eq, PartialOrd, Ord)]
-pub enum Component<'t> {
+pub enum Component<'p> {
     /// The document root
     Root,
-    /// A segment of a JSON Pointer
-    Token(Token<'t>),
-}
-impl<'t> From<Token<'t>> for Component<'t> {
-    fn from(token: Token<'t>) -> Self {
-        Self::Token(token)
-    }
+    /// A token of a [`Pointer`]
+    Token {
+        /// the [`Token`]
+        token: Token<'p>,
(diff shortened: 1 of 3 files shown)


chanced:
I'm setting this PR aside. I'm just going to push 0.5 with the way things are, including `Component` / `Components`. There's room for improvement or removal, not sure which.

I'll leave this one open for now but I don't think `walk_to` / `walk_from` are going to make the cut, at least not with the signature above.

PR #66, opened 2024-08-05
Error on leading zeros when decoding indices

asmello:
In the process of migrating json-patch to version 0.5.1, I noticed one of their test cases was failing because we accept leading zeros and we shouldn't.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -424,6 +424,7 @@ mod tests {
             err.source().unwrap().to_string(),
             "not a number".parse::<usize>().unwrap_err().to_string()
         );
+        assert!(ParseIndexError::LeadingZeros.source().is_none());
     }
 
     #[test]
--- a/src/index.rs
+++ b/src/index.rs
@@ -118,10 +118,10 @@ impl Index {
 
     /// Resolves the index for a given array length.
     ///
-    /// No bound checking will take place. If you wish to ensure the index can
-    /// be used to access an existing element in the array, use [`Self::for_len`]
-    /// - or use [`Self::for_len_incl`] if you wish to accept [`Self::Next`] as
-    /// valid as well.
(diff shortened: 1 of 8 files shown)


src/index.rs
```diff
@@ -260,27 +262,40 @@ impl std::error::Error for OutOfBoundsError {}
...
 /// Indicates that the `Token` could not be parsed as valid RFC 6901 index.
 #[derive(Debug, PartialEq, Eq)]
-pub struct ParseIndexError {
-    /// The source `ParseIntError`
-    pub source: ParseIntError,
+pub enum ParseIndexError {
+    /// The Token does not represent a valid integer.
+    InvalidInteger(ParseIntError),
+    /// The Token contains leading zeros.
+    LeadingZeros,
```
asmello, line 269: This is unfortunately a breaking change, but I don't see another way to represent the additional error condition.

asmello:
@chanced something to look at later: the coverage job is running for each commit, but it should only run in the pull request context. I think some other jobs are also running redundantly.

src/index.rs
chanced: Yea, that needs to go in.

chanced:
@asmello  thanks for the mention, I didn't get a notification for this pull request either. I need to figure out why that is.

chanced:
> coverage :shrug:
😆 

Speaking of which, I really need to go back and clean up the tests at some point.

chanced:
@asmello ugh, I'm sorry. I meant to do a release when I pulled this down but didn't catch that it was aborted. v0.6.0 should be available now.

Screenshot 2024-08-10 at 1 31 35 PM

asmello:
Ha, no problem, haven't gotten around to using it yet. Thanks for noticing!

Issue #31, opened 2024-05-08
Add fluid methods to `PointerBuf`


chanced:
Add fluid methods to `Pointer` that mutate in place and return a new instance.

- with token appended
- with token prepended
- with first token removed (maybe? - dont know whether to send the token into the void)
- with last token removed (same ^^)

this is somewhat obnoxious to deal with:
```rust
let mut instance_location = self.instance_location.clone();
instance_location.push_back(instance.into());
let mut keyword_location = self.keyword_location.clone();
keyword_location.push_back(keyword.into());
```

asmello:
I kind of agree that cloning+push_back isn't super ergonomic. But there isn't a fluent API in std analogs (looking at Path/PathBuf, str/String, slice/Vec), so I'm wondering what's the right design here.

chanced:
@asmello there is, the three are:

- [`with_added_extension`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_added_extension)
- [`with_extension`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_extension)
- [`with_file_name`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_file_name)

asmello:
Yeah, that's a fair point, I missed those. I'm honestly not sure that's great design (it's easy to abuse by chaining multiple times), and the analogy is not perfect: those modify specific parts of the path, while the `Pointer` counterparts extend it. Extension feels like more of a collection operation (and std collections don't have fluent APIs) while modifying a property is more of an struct thing.

This may seem like splitting hairs, but structs have a fixed number of fields, so modifying them in-place tends to not require allocation, while extension is much more likely to. When allocation is likely, being able to reserve ahead of time becomes more important. I think this is why the `Extend` trait works the way it does - it encourages you to do `myvec.extend([1,2,3]);` and not `myvec.push(1).push(2).push(3)`, which is harder to optimise.

chanced:
Right, `extend([1,2,3])` narrows it down to a single alloc vs potentially more.


As an aside, I goofed on some of `Pointerbuf`'s methods btw, like `append` shouldn't return `&Pointerbuf`. I don't know if it'd be worth changing in the future but it caught me off guard when I ran into it. Not sure what the thought process there was.

asmello:
[...]

PR #67, opened 2024-09-25
Adds fluid methods to `Pointer`

chanced:
solves #31 by adding fluid methods to `Pointer`:
- `with_leading_token` - prepends a token
- `with_trailing_token` - appends a token
- `with_appended` - appends a pointer

### DIFF
--- a/src/pointer.rs
+++ b/src/pointer.rs
@@ -437,12 +437,9 @@ impl Pointer {
     /// let foobar = ptr.with_trailing_token("bar");
     /// assert_eq!(foobar, "/foo/bar");
     /// ```
-    pub fn with_trailing_token<'t, T>(&self, token: T) -> PointerBuf
-    where
-        T: Into<Token<'t>>,
-    {
+    pub fn with_trailing_token<'t>(&self, token: impl Into<Token<'t>>) -> PointerBuf {
         let mut buf = self.to_buf();
-        buf.push_back(token);
+        buf.push_back(token.into());
         buf
     }
 
@@ -460,10 +457,7 @@ impl Pointer {
     /// let foobar = ptr.with_leading_token("foo");
     /// assert_eq!(foobar, "/foo/bar");
     /// ```
-    pub fn with_leading_token<'t, T>(&self, token: T) -> PointerBuf
-    where
-        T: Into<Token<'t>>,
-    {
+    pub fn with_leading_token<'t>(&self, token: impl Into<Token<'t>>) -> PointerBuf {
         let mut buf = self.to_buf();
         buf.push_front(token);
         buf
@@ -807,19 +801,13 @@ impl PointerBuf {
     }
 
     /// Pushes a `Token` onto the front of this `Pointer`.
-    pub fn push_front<'t, T>(&mut self, token: T)
-    where
-        T: Into<Token<'t>>,
-    {
+    pub fn push_front<'t>(&mut self, token: impl Into<Token<'t>>) {
         self.0.insert(0, '/');
         self.0.insert_str(1, token.into().encoded());
     }
 
     /// Pushes a `Token` onto the back of this `Pointer`.
-    pub fn push_back<'t, T>(&mut self, token: T)
-    where
-        T: Into<Token<'t>>,
-    {
+    pub fn push_back<'t>(&mut self, token: impl Into<Token<'t>>) {
         self.0.push('/');
         self.0.push_str(token.into().encoded());
     }
@@ -866,10 +854,11 @@ impl PointerBuf {
     ///
     /// ## Errors
     /// A [`ReplaceError`] is returned if the index is out of bounds.
-    pub fn replace<'t, T>(&mut self, index: usize, token: T) -> Result<Option<Token>, ReplaceError>
-    where
-        T: Into<Token<'t>>,
-    {
+    pub fn replace<'t>(
+        &mut self,
+        index: usize,
+        token: impl Into<Token<'t>>,
+    ) -> Result<Option<Token>, ReplaceError> {
         if self.is_root() {
             return Err(ReplaceError {
                 count: self.count(),
--- a/src/pointer.rs
+++ b/src/pointer.rs
@@ -422,6 +422,53 @@ impl Pointer {
     pub fn components(&self) -> Components {
         self.into()
     }
+
+    /// Appends `token` onto `self` as a new [`Pointerbuf`].
+    ///
+    /// ## Example
+    /// ```
+    /// let ptr = jsonptr::Pointer::from_static("/foo");
+    /// let foobar = ptr.with_trailing_token("bar");
+    /// assert_eq!(foobar, "/foo/bar");
+    /// ```
+    pub fn with_trailing_token<'t, T>(&self, token: T) -> PointerBuf
+    where
+        T: Into<Token<'t>>,
+    {
+        let mut buf = self.to_buf();
+        buf.push_back(token);
+        buf
+    }
+
+    /// Prepends `token` onto `self` as a new `PointerBuf`.
+    ///
+    /// ## Example
+    /// ```
+    /// let ptr = jsonptr::Pointer::from_static("/bar");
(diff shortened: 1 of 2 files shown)


chanced:
@asmello do you have any opinion on the names? I struggled with them.

asmello:
I'm not convinced this is the right approach. I feel like `.to_buf()` is something that should always be an explicit step, _especially_ when chaining, as that makes intermediate types less apparent.

Which potentially means the fluent API should exist in `PointerBuf` instead, if at all.

chanced:
This is consistent with `std::path::Path`, e.g. [`Path::with_file_name`](https://doc.rust-lang.org/std/path/struct.Path.html#method.with_file_name)

asmello:
Fair point. I guess paths have some special semantics that make them more like a struct than a collection.

It's kinda interesting, not sure the `std::path::Path` design is very good, as `path.with_extension("foo").with_extension("bar")` results in two allocations, even though it doesn't really need to. We'll be inheriting that quirk. But then you're not really supposed to use this multiple times, so maybe that's ok.

I'm happy with the first two names — they're pretty descriptive, a bit on the long side but I think we don't want to encourage using this API too much, given the potential for allocation abuse, so that's actually good. I think we can do better with the last one though. How about `.concat()`?

chanced:
Yea, hopefully folks won't shoot themselves in the foot with it. Hopefully the ergonomics are worth it and maybe I should add a note about how they allocate and that they above should be avoided.

Oh, I like `concat`. If we go with it, does it make sense to implement [`std::slice::Concat`](https://doc.rust-lang.org/std/slice/trait.Concat.html)?

asmello:
> Maybe I should add a note about how they allocate and that they above should be avoided.

Definitely! This is well documented in the `Path` methods for a reason ;)

If you're happy with `.concat()` I think that's fine. Just wanted to provide another alternative.

> If we go with it, does it make sense to implement [std::slice::Concat](https://doc.rust-lang.org/std/slice/trait.Concat.html)?

I don't think so, for one that trait is nightly and for another its API is different, we'd need to use it like `[ptr1, ptr2].concat()` which doesn't seem very discoverable (or particularly ergonomic).

chanced:
Ah, okay. I wasn't exactly certain how the call looked. From the trait, it seemed straight forward in that it added a `concat` method which returned `Self::Output`. Didn't notice it was nightly either.

I'll rename to `concat` and add docs. This is likely a silly change but I'm dealing with pointers a lot and all mutation is basically clone & then push a token.

Then again.. maybe its not worth the savings in loc.

asmello:
Yeah it's fine to try this out, especially given the `std::path` precedent. The crate is not 1.0 yet so no need to overthink it. Let's just make sure the allocation is well documented and see if we end up making mistakes. 👍

src/pointer.rs
```diff
@@ -422,6 +422,53 @@ impl Pointer {
...
+    ///
+    /// ## Example
+    /// ```
+    /// let ptr = jsonptr::Pointer::from_static("/foo");
+    /// let foobar = ptr.with_trailing_token("bar");
+    /// assert_eq!(foobar, "/foo/bar");
+    /// ```
+    pub fn with_trailing_token<'t, T>(&self, token: T) -> PointerBuf
+    where
+        T: Into<Token<'t>>,
```
asmello, line 436: Minor nit, but I think `impl Into<Token<'_>>` would be cleaner (unsure about whether the lifetime can be elided but I think so)

asmello:
Just a thought, but curious how you start with `&Pointer` and end up having to extend it — could you perhaps start with `PointerBuf` and just reuse it throughout? Quite possible you need a copy, but on the off chance you don't, that'd be an interesting data point.

I'm thinking of some network coding I've been doing where I found that what works best is to allocate a single permanent buffer at a high level, sift it down the abstraction layers up to the point where I do a socket read, and then on the way back up all the parsing (well, most of it) can be done with zero copies. It's quite neat from a lifetime management perspective, but most importantly it helps monstrously with reducing allocations. It's quite easy and tempting to allocate a fresh buffer for every read but that's also easily ends up being a bottleneck. The `Buf` trait also helps do that while allowing the buffer to be dynamically extended as needed (eg, [`UdpSocket::recv_buf`](https://docs.rs/tokio/latest/tokio/net/struct.UdpSocket.html#method.recv_buf)).

chanced:
Nah, I can't reuse the buffer as they get stored.

This is a function that resolves a source and inserts it into my source database :

```rust
async fn resolve_path<'int, R: 'static + Resolve>(
    sources: &'int mut Sources,
    resolver: &R,
    uri: &AbsoluteUri,
    base: AbsoluteUri,
    fragment: String,
) -> Result<Resolved, Error<R>> {
    let resolved_root = resolve_root(sources, resolver, &base).await?;
    let located = resolved_root.located;
    let document_key = resolved_root.document_key;

    let relative_path = Pointer::parse(&fragment).map_err(|source| Error::InvalidPointer {
        uri: uri.clone(),
        source,
    })?;


    let mut absolute_path = sources.source(resolved_root.source_key).path().to_buf(); // <-- pointer related stuff
    absolute_path.append(&relative_path);                                                                            // <--

    let source_key = sources
        .link(New {
            uri: uri.clone(),
            document_key,
            absolute_path,
            fragment: Some(Fragment::Pointer(relative_path.to_buf())),
        })
        .map_err(|source| match source {
            LinkError::InvalidPath(err) => Error::PathNotFound {
                uri: uri.clone(),
                source: err,
            },
            LinkError::SourceConflict(err) => unreachable!(
                "\n\nsource conflict:\n\nuri: {uri}\ndocument_key: {document_key:?}\nerr: {err}",
            ),
        })?;

    Ok(Resolved::Source(resolved::Src {
        link_created: true,
        source_key,
        document_key,
        located,
    }))
}
```
https://github.com/chanced/grill/blob/0d3e2bfc4505d5638717b55826bf8ae6c2a5fc83/grill-json-schema/src/compile/scan/resolve.rs#L99-L139

asmello:
Well, fair enough. Was worth a shot.

chanced:
Absolutely, and I appreciate it.

I updated the docs to reflect std with an additional note that may be a bit too much.

chanced:
@asmello I tried using the inline, `t: impl` syntax but I'm running into type ambiguity with it:

```rust
    pub fn push_back<'t, T>(&mut self, token: impl Into<Token<'t>>) {
        self.0.push('/');
        self.0.push_str(token.into().encoded());
    }
    pub fn with_trailing_token<'t>(&self, token: impl Into<Token<'t>>) -> PointerBuf {
        let mut buf = self.to_buf();
        buf.push_back(token.into());
        buf
    }
```

```bash
▶ cargo check
    Checking jsonptr v0.6.0 (/Users/chance/dev/jsonptr)
error[E0282]: type annotations needed
   --> src/pointer.rs:442:13
    |
442 |         buf.push_back(token.into());
    |             ^^^^^^^^^ cannot infer type for type parameter `T` declared on the method `push_back`

error[E0282]: type annotations needed
   --> src/pointer.rs:462:13
    |
462 |         buf.push_front(token);
    |             ^^^^^^^^^^ cannot infer type for type parameter `T` declared on the method `push_front`

For more information about this error, try `rustc --explain E0282`.

File: src/pointer.rs [hunk 2]
```diff
@@ -799,10 +854,11 @@ impl PointerBuf {
...
     /// A [`ReplaceError`] is returned if the index is out of bounds.
-    pub fn replace<'t, T>(&mut self, index: usize, token: T) -> Result<Option<Token>, ReplaceError>
-    where
-        T: Into<Token<'t>>,
-    {
+    pub fn replace<'t>(
+        &mut self,
+        index: usize,
+        token: impl Into<Token<'t>>,
+    ) -> Result<Option<Token>, ReplaceError> {
```
asmello, line 861: Ah, nice one!

asmello (approved):

chanced:
@asmello thanks for the help & review.

PR #69, opened 2024-09-30
adds `len` and `is_empty` to `Pointer`

chanced:
Adds `len` and `is_empty` to `Pointer`.

### DIFF
--- a/src/pointer.rs
+++ b/src/pointer.rs
@@ -475,13 +475,45 @@ impl Pointer {
     /// ## Examples
     /// ```
     /// let ptr = jsonptr::Pointer::from_static("/foo");
-    /// let barbaz = jsonptr::Pointer::from_static("/bar/baz");
-    /// assert_eq!(ptr.concat(&barbaz), "/foo/bar/baz");
+    /// let other = jsonptr::Pointer::from_static("/bar/baz");
+    /// assert_eq!(ptr.concat(other), "/foo/bar/baz");
     /// ```
     pub fn concat(&self, other: &Pointer) -> PointerBuf {
         let mut buf = self.to_buf();
         buf.append(other);
         buf
+    }
+
+    //  Returns the length of `self` in encoded format.
+    ///
+    /// This length is in bytes, not [`char`]s or graphemes. In other words, it might
+    /// not be what a human considers the length of the string.
+    ///
+    /// ## Examples
+    /// ```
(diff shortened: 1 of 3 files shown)


chanced:
@asmello we had talked about adding `len` before so I'm just going to go ahead and merge this.

src/pointer.rs
```diff
@@ -463,7 +463,7 @@ impl Pointer {
         buf
     }
 
-    /// Creates an owned [`Pointerbuf`] like `self` but with `other` appended to
+    /// Creates an owned [`PointerBuf`] like `self` but with `other` appended to
```
asmello, line 466: Had to look like 5x to notice what changed, omg.

src/pointer.rs
```diff
@@ -475,14 +475,46 @@ impl Pointer {
...
+    /// ## Examples
+    /// ```
+    /// let mut ptr = jsonptr::Pointer::from_static("/foo/bar").to_buf();
+    /// assert_eq!(ptr.len(), 8);
+    ///
+    /// ptr.push_back("~");
+    /// assert_eq!(ptr.len(), 11);
+    ///
+    /// ```
+    pub fn len(&self) -> usize {
```
asmello, line 501: I think this is sane. One can always do `ptr.tokens().count()` for the alternative definition of "length".

I see that you've copied the doc-comment from `str::len`. This is mostly accurate, since we do indeed just delegate to that method, but the semantics is not quite right. I think we could go with something simpler instead, along the lines of:

```
    /// This length expresses the byte count of the underlying string that represents the RFC 6091 Pointer. See also [`std::str::len`].
```

Although the caveat with char/graphemes applies, it's less important for us than the relationship with `std::str`.

asmello (commented):
Leaving a comment on docs we can fix later.

chanced:
Yea, that makes sense.

chanced:
Ugh, I accidentally added them to both `Pointer` and `PointerBuf`.

It would be a breaking change to delete them from `PointerBuf`, right?

asmello:
> Ugh, I accidentally added them to both `Pointer` and `PointerBuf`.
> 
> It would be a breaking change to delete them from `PointerBuf`, right?

I did wonder. I don't think it is, because `PointerBuf` implements `Deref<Target=Pointer>`, so `ptr.len()` works regardless.

PR #72, opened 2024-09-30
improves `len`docs

chanced:
fixes #72

### DIFF
--- a/src/pointer.rs
+++ b/src/pointer.rs
@@ -486,8 +486,8 @@ impl Pointer {
 
     //  Returns the length of `self` in encoded format.
     ///
-    /// This length is in bytes, not [`char`]s or graphemes. In other words, it might
-    /// not be what a human considers the length of the string.
+    /// This length expresses the byte count of the underlying string that
+    /// represents the RFC 6091 Pointer. See also [`std::str::len`].
     ///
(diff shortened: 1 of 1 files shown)


chanced:
> I think this is sane. One can always do `ptr.tokens().count()` for the alternative definition of "length".
>
> I see that you've copied the doc-comment from `str::len`. This is mostly accurate, since we do indeed just delegate to that method, but the semantics is not quite right. I think we could go with something simpler instead, along the lines of:
>
> 
>    /// This length expresses the byte count of the underlying string that represents the RFC 6091 Pointer. See also [`std::str::len`].
>
> Although the caveat with char/graphemes applies, it's less important for us than the relationship with `std::str`.

_Originally posted by @asmello in https://github.com/chanced/jsonptr/pull/69#discussion_r1781566911_

Carry over from the above.

I guess I leave `len` and `is_empty` on `PointerBuf` for now, right? Ugh. I realized I had accidentally added them to `PointerBuf` and I thought I deleted them to move them on to `Pointer` but I guess I yanked instead :(.

chanced:
I yanked v06.2 - i seriously doubt it'd be a breaking change for anyone but I wanted to make sure since the CI reported it as such.

PR #76, opened 2024-10-14
Refactor constructor safety

asmello:
Although a lot more inconvenient for us, some of our functions have invariants that have to be held externally, so they should be `unsafe fn`. This also makes it justifiable to expose them publicly (as then users can opt-in for manual checking of the invariants).

### Notable changes:
- Expose `Pointer::new` as a public (but unsafe) constructor (now called `Pointer::new_unchecked`). This gives users a way to bypass validation if they're sure they've got a valid string. Akin to `String::from_utf8_unchecked` and family.
- Add `PointerBuf::new_unchecked` as a new unsafe constructor, to match `Pointer::new_unchecked`.
- Made `Token::from_encoded_unchecked` unsafe. This one I didn't make public because I don't see much use for it. We may revisit this later.

### DIFF
--- a/src/pointer.rs
+++ b/src/pointer.rs
@@ -542,7 +542,7 @@ impl Pointer {
     ///
     /// ## Examples
     /// ```
-    /// let mut ptr = jsonptr::PointerBuf::root();
+    /// let mut ptr = jsonptr::PointerBuf::new();
     /// assert!(ptr.is_empty());
     ///
     /// ptr.push_back("foo");
@@ -861,10 +861,8 @@ impl PointerBuf {
     ///
     /// ## Errors
     /// Returns a [`ParseError`] if the string is not a valid JSON Pointer.
-    pub fn parse(s: impl Into<String>) -> Result<Self, ParseError> {
-        let s = s.into();
-        validate(&s)?;
-        Ok(Self(s))
+    pub fn parse<S: AsRef<str> + ?Sized>(s: &S) -> Result<Self, ParseError> {
+        Pointer::parse(&s).map(Pointer::to_buf)
     }
 
     /// Creates a new `PointerBuf` from a slice of non-encoded strings.
@@ -1509,7 +1507,7 @@ mod tests {
             assert_eq!(ptr, Pointer::root());
         }
         {
-            let mut ptr = PointerBuf::root();
+            let mut ptr = PointerBuf::new();
             assert_eq!(ptr.tokens().count(), 0);
             ptr.push_back("");
             assert_eq!(ptr.tokens().count(), 1);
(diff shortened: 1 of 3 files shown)


asmello:
I think the sanitizer test is failing due to a bug in the sanitizer itself (given the obscure error referencing its source code).

chanced:
I'm at the fair with my family but noticed this come through and just wanted to say that I'm stoked about this one. I considered suggesting exposing Pointer::new but was hesitant.

I'll review both tomorrow morning. 

Also, you're awesome dude.

asmello:
No hush, just came with a few ideas recently.

I noticed `json-patch` recently upgraded to 0.6.0, which is great (and I feel bad I let them do it, meant to contribute a PR myself).

chanced:
Hah, you did! Perhaps not to hatch directly but the state of this crate would be in much worse shape without your contributions. You brought it up a few levels! :)

src/pointer.rs
```diff
@@ -49,11 +49,14 @@ impl core::fmt::Display for Pointer {
...
-    /// Private constructor for strings that are known to be correctly encoded
+    /// Create a `Pointer` from a string that is known to be correctly encoded.
     ///
-    /// This is a zero-copy constructor
-    fn new<S: AsRef<str> + ?Sized>(s: &S) -> &Self {
-        unsafe { &*(core::ptr::from_ref::<str>(s.as_ref()) as *const Self) }
+    /// This is a cost-free conversion.
+    ///
+    /// ## Safety
+    /// The provided string must adhere to RFC 6901.
```
chanced, line 57:
```suggestion
    /// The provided string must adhere to [RFC 6901](https://datatracker.ietf.org/doc/html/rfc6901):
    /// - The pointer must start with `'/'` (%x2F) unless empty
    /// - Tokens must be properly encoded:
    ///     - `'~'` (%x7E) must be escaped as `"~0"`
    ///     - `'/'` (%x2F) must be escaped as `"~1"`
    /// For potentially fallible parsing, see [`Pointer::parse`].
```
I think being explicit here about what's expected may help some folks - or at least persuade them to use something else :). Realistically though, `unsafe` will likely make folks take notice.

chanced (commented):
Ah, you went above and beyond with comments in this one :)

Looks great to me!

src/pointer.rs
asmello: Yeah, this seems sensible to me, thanks for writing it down.

asmello:
To be honest the unsafe for internal uses might be a bit overkill, I'm not 100% sure this is a great idea long term, but since it's an internal detail I think it's ok to experiment with.

PR #77, opened 2024-10-14
update signature of PathBuf::parse

asmello:
This is what it should've always been. It's technically a breaking change, but I think it's ok because most types that implement `AsRef<str>` would also implement `Into<String>`.


chanced:
Yep, this is definitely what it should have been all along - oops!

Even though it is almost certainly a non-breaking change for the vast majority of codebases, we still need a minor/breaking version bump here, right?

asmello:
> Even though it is almost certainly a non-breaking change for the vast majority of codebases, we still need a minor/breaking version bump here, right?

Yep, that's the responsible thing to do.

chanced:
I just realized the version bump is going to need a minor as well. It sounds like we should just wrap all 3 PRs into one minor release.

PR #78, opened 2024-10-14
use associated type bound

asmello:
[See RFC](https://rust-lang.github.io/rfcs/2289-associated-type-bounds.html). This was stabilized in 1.79.0.


asmello:
I see that we declare 1.76 in the manifest. Shall we bump that now? Or is it too early still?

PR #79, opened 2024-10-14
Add some box-related methods

asmello:
May come in handy sometimes.


chanced:
This one looks good. I'm going to merge this one and you already merged the other.

Are you cool with cutting a patch with these and then a minor with the last remaining?

PR #82, opened 2024-10-19
Add `Token::is_next` method

asmello:
Small QoL addition.

Also open to naming it `is_next_idx` to avoid ambiguity if used in combination with `Pointer::tokens` (I don't think this is particularly confusing, but 🤷 ).


chanced:
lgtm

Issue #42, opened 2024-06-20
Allow for splitting a `Pointer` by `Token` position (index)


chanced:
We need a way to slice / split a token by range, using the index.

```text
/foo/bar/baz
↑   ↑   ↑
0   1   2
```

chanced:
Alternatively, we just continue to use the offsets.

chanced:
Dealing with `split_at` and using offsets is somewhat a pain in the ass to deal with.

I'm still not sure what the right answer is here. Some of the API treats the `Pointer` as a string, and in a lot of cases this is what you'd want, but it also behaves like a collection.

If we double down on the `Pointer` being a collection, then we'd need to re-evaluate the methods which treat it as a string.

chanced:
It wasn't _horrible_ to write the iterator below, but I think it could be made a lot cleaner. If `Component` were something like

```rust
enum Component<'a> {
    Root,    
    Component {
        /// full path to this component
        path: &'a Pointer, 
        offset: usize,
        token: Token<'a>,
    }
}
```

Here's the iterator:
```rust
#[derive(Debug)]
pub struct WalkTo<'p, 'v> {
    cursor: &'v Value,
    original_value: &'v Value,
    full_path: &'p Pointer,
    offset: Option<NonZeroUsize>,
}
impl<'p, 'v> WalkTo<'p, 'v> {
    pub fn new(value: &'v Value, to: &'p Pointer) -> Self {
        Self {
            cursor: value,
            original_value: value,
            full_path: to,
            offset: None,
        }
    }

    fn handle_err(
        &self,
        _err: ResolveError,
        path: &'p Pointer,
        _resolvable: &'p Pointer,
    ) -> Option<Result<(&'p Pointer, &'v Value), ResolveError>> {
        // TODO: determine appropriate offsets and update error rather than re-resolving
        // for now, we are just punting to `Resolve` to recreate the error
        Some(Err(self.original_value.resolve(path).unwrap_err()))
    }
}

impl<'p, 'v> Iterator for WalkTo<'p, 'v> {
    type Item = Result<(&'p Pointer, &'v Value), ResolveError>;
    fn next(&mut self) -> Option<Self::Item> {
        // we need to get the offset to determine where we are in the pointer
        let offset = if let Some(offset) = self.offset {
            // the offset has previously been set, so we use that
            offset.get()
        } else {
            // An empty offset means we are at the beginning of the walk.

            // this is a special case where the target path is root
[...]

PR #83, opened 2024-10-20
Proposal for `Walk` / `WalkMut`

chanced:
The 
```rust
pub trait Walk: Resolve<Error: Debug> {
    type WalkTo<'value, 'ptr>;    
    type WalkFrom<'value>;
    fn walk_to<'v, 'p>(&'v self, to: &'p Pointer) -> Self::WalkTo<'v, 'p>;
    fn walk_from(&self, from: PointerBuf) -> Result<Self::WalkFrom<'_>, Self::Error>;
    fn walk(&self) -> Self::WalkFrom<'_> {
        self.walk_from(PointerBuf::default()).unwrap()
    }
}
```

```rust
pub trait WalkMut: ResolveMut<Error: Debug> {
    type WalkTo<'value, 'ptr>;    
    type WalkFrom<'value>;
    fn walk_to_mut<'v, 'p>(&'v mut self, to: &'p Pointer) -> Self::WalkTo<'v, 'p>;
    fn walk_from_mut(&mut self, from: PointerBuf) -> Result<Self::WalkFrom<'_>, Self::Error>;
    fn walk_mut(&mut self) -> Self::WalkFrom<'_> {
        self.walk_from_mut(PointerBuf::default()).unwrap()
    }
}
```

Each iterator returns the full path of the value along with a (`mut`) reference to the current node. 

In the case of `WalkTo`, the iterator follows the path of the `Pointer`, emitting the result of `resolve` of each token as if the full path were resolved.

For `WalkFrom`, it walks a `Value` from a given `Pointer` onward, emitting the full path (as a `PathBuf`) for each node.

I have `WalkTo` implemented. `WalkFrom` I have a previous implementation specific to JSON that I can make generic. I haven't bothered with `mut` of either.

This could be far too niche to bother. As such I'll hold off on finishing it.

@asmello let me know what you think.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,9 +27,10 @@ syn = { version = "1.0.109", optional = true }
 
 [features]
 assign  = []
-default = ["std", "serde", "json", "resolve", "assign", "delete"]
+default = ["std", "serde", "json", "resolve", "assign", "delete", "walk"]
 delete  = ["resolve"]
 json    = ["dep:serde_json", "serde"]
 resolve = []
 std     = ["serde/std", "serde_json?/std"]
 toml    = ["dep:toml", "serde", "std"]
+walk    = ["resolve"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -64,6 +64,9 @@ pub mod resolve;
 #[cfg(feature = "resolve")]
 pub use resolve::{Resolve, ResolveMut};
 
+#[cfg(feature = "walk")]
+pub mod walk;
+
 mod pointer;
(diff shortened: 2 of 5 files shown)


chanced:
as discussed in #42

chanced:
As of right now, this doesn't seem like the right abstractions for this crate.

I'm going to leave this open for now incase any users would like to chime in on ideas or needs for similar abstractions.

chanced:
On second thought, this is about to become completely obsolete with `get` now accepting ranges. Any iterator built with similar intent from here on out should be a lot cleaner.

PR #84, opened 2024-10-20
Slicing API for `Pointer`

asmello:
Follow up of some ideas in #42.

This is effectively an extension of the split methods that can slice and dice `Pointer` types without allocation, similar to how `slice::get` works.

We use token indices as the range bounds, which ensures that any valid bound produces a valid `&Pointer` as a result.

This is not a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 -   Changed signature of `PathBuf::parse` to avoid requiring allocation.
 -   Bumps minimum Rust version to 1.79.
+-   `Pointer::get` now accepts ranges and can produce `Pointer` segments as output (similar to
+    `slice::get`).
 
 ## [0.6.2] 2024-09-30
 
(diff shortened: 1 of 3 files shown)


chanced:
SliceIndex did work!? That's awesome! You are damn good at this man.

I'm still on mobile, out and about, but I'll look at it tonight, hopefully.

chanced:
Man, this language is awesome.

This solves #42. We already have a way to get a token by index, this allows for the rest (splitting, ranging, etc). Moving forward, we can reference tokens by index or range if we ever add additional nice-to-have methods.

#42 covered more than the range or being able to split at tokens. I'm going to split off a couple more tickets from it and reduce it down to simply being able to split at an index.

asmello:
> SliceIndex did work!? That's awesome! You are damn good at this man.

Nope, it didn't! You weren't wrong about that remark, it's just that we're not actually meant to implement `SliceIndex` (it's even a sealed trait, for good reason). I introduced our own trait instead.

> Man, this language is awesome.

I tend to agree, it's pretty neat I was able to express what I wanted with the trait system.

asmello:
Looking again I think `SlicePointer` is a bit awkward as a name, I meant it as "you slice a pointer with it", but that's probably not how most people will read it. `PointerSlice` probably makes more sense. Will follow up with a quick rename before we release.

asmello:
Actually, I'm thinking about it all wrong. It's `SliceIndex`, so it should be `PointerIndex`, duh.

PR #85, opened 2024-10-20
add wngr as author

asmello:
@wngr

### DIFF
--- a/.taplo.toml
+++ b/.taplo.toml
@@ -1,24 +1,24 @@
 include = ["**/*.toml", "**/Cargo.toml", "Cargo.toml"]
 [formatting]
-  align_comments      = true
-  align_entries       = true
-  allowed_blank_lines = 1
-  indent_entries      = false
-  reorder_arrays      = false
-  reorder_keys        = true
-  trailing_newline    = true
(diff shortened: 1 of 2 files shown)


asmello:
We should probably have a CI check for the manifest formatting

asmello:
Interesting, looks like adding the extra key makes a big change in how taplo chooses to format the file. This is possibly a side-effect of the custom settings in `.taplo.toml`. Might be something we wish to correct (but then again, changes to the TOML files aren't that frequent so maybe this is ok).

chanced:
Yea, taplo isn't always consistent with formatting. Sometimes it triggers and sometimes it just flakes out. 

I'll circle back and look at the config later this evening.

chanced:
@wngr just want to make sure you're cool with this before I pull it in.

chanced:
In case it wasn't obvious (it may not be) - I very much believe you should be on the authors list. 

I just want to make sure you don't have oppositions for reasons.

wngr:
Appreciate your considerations -- happy to be included! :-)

chanced:
Rock on :)

PR #86, opened 2024-10-20
check toml fmt

asmello:
Noticed formatting wasn't being enforced in #85


asmello:
Actually, let me try from a branch so I can actually test the workflow.

PR #87, opened 2024-10-20
check toml fmt

asmello:
Noticed formatting wasn't being enforced in #85


asmello:
Ok, now we're cooking. Will apply the formatting and let's merge this before #85 to minimise the diff.

Issue #88, opened 2024-10-21
Deprecate `Pointer::split_at`


chanced:
We need to deprecate `Pointer::split_at` so that it can eventually be replaced by a method of the same name that utilizes position/index rather than offset.

PR #89, opened 2024-10-21
Deprecates `Pointer::split_at`, adds `Pointer::split_at_offset`

chanced:
Solves #88 

- Deprecates `Pointer::split_at`
- Adds `Pointer::split_at_offset`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-
 ## [Unreleased]
 
 ### Added
 
--   Adds method `into_buf` for `Box<Pointer>` and `impl From<PathBuf> for Box<Pointer>`. 
+-   Adds method `into_buf` for `Box<Pointer>` and `impl From<PathBuf> for Box<Pointer>`.
 -   Adds unsafe associated methods `Pointer::new_unchecked` and `PointerBuf::new_unchecked` for
     external zero-cost construction.
 -   Adds `Pointer::starts_with` and `Pointer::ends_with` for prefix and suffix matching.
 -   Adds new `ParseIndexError` variant to express the presence non-digit characters in the token.
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
+-   Adds `Pointer::split_at_offset` to replace the deprecated `Pointer::split_at`.
 
 ### Changed
 
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Bumps minimum Rust version to 1.79.
 -   `Pointer::get` now accepts ranges and can produce `Pointer` segments as output (similar to
     `slice::get`).
+-   Deprecates `Pointer::split_at`
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


src/pointer.rs
```diff
@@ -233,7 +233,37 @@ impl Pointer {
     /// assert_eq!(tail, Pointer::from_static("/bar/baz"));
     /// assert_eq!(ptr.split_at(3), None);
     /// ```
+    #[deprecated(
+        since = "0.8.0",
```
asmello, line 237: Did we do 0.7 already?

src/pointer.rs
```diff
@@ -233,7 +233,37 @@ impl Pointer {
...
+    ///
+    /// ```rust
+    /// # use jsonptr::Pointer;
+    /// let ptr = Pointer::from_static("/foo/bar/baz");
+    /// let (head, tail) = ptr.split_at(4).unwrap();
+    /// assert_eq!(head, Pointer::from_static("/foo"));
+    /// assert_eq!(tail, Pointer::from_static("/bar/baz"));
+    /// assert_eq!(ptr.split_at_offset(3), None);
+    /// ```
+    pub fn split_at_offset(&self, offset: usize) -> Option<(&Self, &Self)> {
```
asmello, line 266: This one is safe because it returns an `Option`, but should we maybe make it unsafe (and not check the boundary condition)?

I imagine that one either will have a valid offset to use already, or they'll use `split_at` instead. Seems unlikely they'll try some offset they're unsure is correct.

CHANGELOG.md
```diff
@@ -5,24 +5,25 @@ All notable changes to this project will be documented in this file.
...
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
+-   Adds `Pointer::split_at_offset` to replace the deprecated `Pointer::split_at`.
 
 ### Changed
 
 -   Changed signature of `PathBuf::parse` to avoid requiring allocation.
 -   Bumps minimum Rust version to 1.79.
 -   `Pointer::get` now accepts ranges and can produce `Pointer` segments as output (similar to
     `slice::get`).
+-   Deprecates `Pointer::split_at`
```
asmello, line 26: `Deprecated` can have [its own section](https://keepachangelog.com/en/1.1.0/#how).

src/pointer.rs
chanced: I kept the signature the same as `split_at` assuming we'd deprecate it and phase it out. Then perhaps re-introduce it using the position instead.

I'm truly not sure why I opted to go for `offset` in hindsight. It's like I spaced on the fact that I had `get` etc that were index based. Rather annoying mistake on my part.

chanced: What about adding `split_at_offset_unchecked`? The reason I'm hesitant is solely for crates that avoid `unsafe` entirely.

asmello: > I'm truly not sure why I opted to go for offset in hindsight. It's like I spaced on the fact that I had get etc that were index based. Rather annoying mistake on my part.

Don't beat yourself over it, it wasn't as obvious back then. I also didn't realise we could extend the `get` method to achieve what we really wanted to do with that; if I had, I'd have pushed back more.

> What about adding split_at_offset_unchecked? The reason I'm hesitant is solely for crates that avoid unsafe entirely.

Those can opt to use `split_at` ;)

Unless your concern is over a lack of alternative in the interim? I suppose we can also keep `split_at_offset` with the current signature, though we'd probably deprecate it too longer term. Feels a bit odd to add something we may remove later, but I think it makes sense in this case.

chanced: Hmm. Good points. You're right that it is odd - especially if we don't end up spacing the releases.

src/pointer.rs
chanced: Lol, no. I skipped ahead. Thanks!

asmello:
Are we still moving forward with this @chanced?

PR #94, opened 2024-10-21
Improve `ParseIndexError::InvalidCharacter` error

asmello:
In #80 I introduced a new error variant to capture cases where the array index starts with `+` (which is valid for `usize` representations in Rust, but not valid for array indices per RFC 6901).

As I noted [here](https://github.com/chanced/jsonptr/pull/80#issuecomment-2427119055), keeping a collection of invalid characters in the error variant is overkill, since the accepted domain of `usize::parse` is _very_ unlikely to evolve, as that'd be a breaking change in `std`. According to [the docs](https://doc.rust-lang.org/std/primitive.usize.html#method.from_str_radix) `+` is the only accepted non-digit character, so we don't need to expect any others.

I'm not specialising the variant name for `+` though to leave some room for later adopting a different parser that has different domain (unlikely as that is) or in the off chance `std` does change this API. The error message changing isn't a hard break, if we ever need to update it in the future, so I'm keeping it more specific to make it more useful.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -166,15 +166,22 @@ impl FromStr for Index {
         } else if s.starts_with('0') && s != "0" {
             Err(ParseIndexError::LeadingZeros)
         } else {
-            let idx = s.parse::<usize>().map(Index::Num)?;
-            if s.chars().all(|c| c.is_ascii_digit()) {
-                Ok(idx)
-            } else {
-                // this comes up with the `+` sign which is valid for
-                // representing a `usize` but not allowed in RFC 6901 array
-                // indices
-                Err(ParseIndexError::InvalidCharacter)
-            }
+            s.chars().position(|c| !c.is_ascii_digit()).map_or_else(
+                || {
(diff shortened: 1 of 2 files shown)


asmello:
I wonder if we should also make the validation more specific (to reduce the chance of the error message lying to the user), at the risk of accepting an invalid input we didn't expect to pass the `usize::parse` parsing. Probably overthinking it.

src/index.rs
```diff
@@ -166,21 +166,22 @@ impl FromStr for Index {
...
-                invalid.dedup();
-                Err(ParseIndexError::InvalidCharacters(
-                    invalid.into_iter().collect(),
-                ))
-            }
+            s.chars().position(|c| !c.is_ascii_digit()).map_or_else(
+                || {
+                    s.parse::<usize>()
+                        .map(Index::Num)
+                        .map_err(ParseIndexError::from)
```
asmello, line 173: This should almost never occur now, the only case I think it might is if the number is too large.

PR #95, opened 2024-10-22
Refactor errors to include position, span (offset + len), and optional miette integration

chanced:
This pull request:
- Adds `Span` structure which contains `len` & `offset` 
- Adds `Spanned` error wrapper struct that contains a `Span` and a source error. It implements `miette::Diagnostic` by acting as a pass-through for all methods except `labels` (used to mark spans). For `labels`, it provides a blank (`None`) label with the span
- Adds `Positioned` error wrapper struct that contains a `Spanned` error and a position. It also implements `miette::Diagnostic` by deferring to `Spanned`
- Changes all errors to tuple variants, using either `Positioned` or `Spanned`, depending on context.

I'm just getting started with it so it will not compile right now. I should hopefully have it done today.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,6 +57,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1e087f84d4f86bf4b218b927129862374b72199ae7d8657835f1e89000eea4fb"
 
 [[package]]
+name = "higher-kinded-types"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "561985554c8b8d4808605c90a5f1979cc6c31a5d20b78465cd59501233c6678e"
+dependencies = [
+ "never-say-never",
+]
+
+[[package]]
 name = "indexmap"
 version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -77,6 +86,7 @@ name = "jsonptr"
 version = "0.6.3"
 dependencies = [
  "miette",
+ "polonius-the-crab",
  "quickcheck",
  "quickcheck_macros",
  "serde",
@@ -133,6 +143,22 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.82",
+]
+
+[[package]]
+name = "never-say-never"
+version = "6.6.666"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf5a574dadd7941adeaa71823ecba5e28331b8313fb2e1c6a5c7e5981ea53ad6"
+
+[[package]]
+name = "polonius-the-crab"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d9b0ba45e2c294fcd0795bc51e45c12746758e9954bb1190f97603ec9a3e91"
+dependencies = [
+ "higher-kinded-types",
+ "never-say-never",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d4c819a1287eb618df47cc647173c5c4c66ba19d888a6e50d605672aed3140de"
 
 [[package]]
+name = "cfg-if"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
 name = "env_logger"
 version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -39,22 +45,22 @@ version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ee8025cf36f917e6a52cce185b7c7177689b838b7ec138364e50cc2277a56cf4"
 dependencies = [
- "cfg-if",
+ "cfg-if 0.1.2",
  "libc",
  "wasi",
 ]
 
 [[package]]
 name = "hashbrown"
-version = "0.14.0"
+version = "0.15.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2c6201b9ff9fd90a5a3bac2e56a830d0caa509576f0e503818ee82c181b3437a"
+checksum = "1e087f84d4f86bf4b218b927129862374b72199ae7d8657835f1e89000eea4fb"
 
 [[package]]
 name = "indexmap"
-version = "2.0.0"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d5477fe2230a79769d8dc68e0eabf5437907c0457a5614a9e8dddb67f65eb65d"
+checksum = "707907fe3c25f5424cce2cb7e1cbcafee6bdbe735ca90ef77c29e84591e5b9da"
 dependencies = [
  "equivalent",
  "hashbrown",
@@ -70,6 +76,7 @@ checksum = "d2f3e61cf687687b30c9e6ddf0fc36cf15f035e66d491e6da968fa49ffa9a378"
 name = "jsonptr"
 version = "0.6.3"
 dependencies = [
+ "miette",
  "quickcheck",
  "quickcheck_macros",
  "serde",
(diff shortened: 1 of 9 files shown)


chanced:
@asmello incase you happen to have a minute - can you spot this and give me a bit of feedback on how you feel about both the degree of breaking change this is going to introduce as well as the general design? I only spot checked on impl of `miette` and it was kinda convoluted.

Here's a short summary, told through types:

```rust
pub struct Span {
    pub offset: usize,
    pub len: usize,
}

pub struct Spanned<E> {
    span: Span,
    source: E,
}

pub struct Positioned<E> {
    position: usize,
    source: Spanned<E>,
}

pub enum Error {
    FailedToParseIndex(Positioned<ParseIndexError>),
    OutOfBounds(Positioned<OutOfBoundsError>),
}

#[cfg(feature = "miette")]
impl<E> miette::Diagnostic for Spanned<E>
where
    E: 'static + miette::Diagnostic,
{
  // passthrough to `self::source`
    fn labels(&self) -> Option<Box<dyn Iterator<Item = miette::LabeledSpan> + '_>> {
        Some(Box::new(once(miette::LabeledSpan::new(
            None,
            self.offset(),
            self.len(),
        ))))
    }
}


#[cfg(feature = "miette")]
impl<E> miette::Diagnostic for Positioned<E>
where
    E: 'static + miette::Diagnostic,
{
  // passthrough to `Spanned`
}
```

asmello:
I haven't used `miette` like this (my uses have been through the dynamic macros or `thiserror` integration), but will give a look.

chanced:
I will need to use the macro on the errors themselves (e.g. `ParseIndexError`)

asmello:
Oh I think I understand, each `Error` will implement `Diagnostic` on its own, but we'll leave the labels to a wrapper since that part is common to most (all?) of them. I think this makes sense. We may want some wrapper for the source code part too, for a similar benefit. After all the labels are pointless without the source code.

I think it'll be worth playing around with a few errors and seeing how it goes. Biggest risk I see is this ends up being equivalent work to just implementing these directly for each error type.

chanced:
Oh, right! I forgot about source!

So the reason I opted for the wrappers is that most of the errors themselves do not have knowledge of their surroundings. I can change that though with wrappers specific to each. 

```rust
/// Indicates that an `Index` is not within the given bounds.
#[derive(Debug, PartialEq, Eq)]
pub struct OutOfBoundsError {
    pub length: usize,
    pub index: usize,
}
```
to 
```rust
#[derive(Debug, PartialEq, Eq)]
pub struct OutOfBoundsErrorWrapper {
    pub source: OutOfBoundsError,
    pub position: usize,
   pub span: Span,
}
```
or something like it

chanced:
Or I guess I could go the other route and refactor the error into a `struct`:

```rust
// assign::Error
pub struct Error {
     kind: Kind,
     position: usize,
     span: Span,
}

pub enum Kind {
    FailedToParseIndex(ParseIndexError),
    OutOfBounds(OutOfBoundsError),
}
```

chanced:
I'm going to flush this design out to see how it works - I haven't figured out what to do about source yet (the convoluted nature of `pest`'s errors now make a lot more sense) so it may be all for not.

chanced:
I have a general idea for how to do this - just running into borrowck issues. I have no idea how the ergonomics are going to shake out. The hope is this:

```rust
let mut buf = PointerBuf::parse("/example"); // 'a
let err_report: Report<'a, ReplaceError, Pointer> = buf.report_err().replace(1, "oops").unwrap_err();
```

The way it is anticipated to work is:

```rust
// crate::error::Reporter
pub trait Reporter<T> {
    type Reporter<'e>
    where
        Self: 'e;

    // TODO: naming, alts: report, include_err_report, with_err_report, ???
    fn report_err(&'_ self) -> Self::Reporter<'_>;
}

```rust
// crate::error::MutReporter
pub trait MutReporter<T> {
    type Reporter<'e>
    where
        Self: 'e;

    // TODO: naming, alts: report, include_err_report, with_err_report, ???
    fn report_err(&'_ mut self) -> Self::Reporter<'_>;
}


pub struct MutReporter<'p> {
    ptr: &'p mut PointerBuf,
}

impl<'p> MutReporter<'p> {
    pub fn replace<'t>(
        mut self,
        index: usize,
        token: impl Into<Token<'t>>,
    ) -> Result<Option<Token<'t>>, Report<'p, ReplaceError, Pointer>>
    where
        'p: 't,
    {
        match self.ptr.replace(index, token) {
            Ok(res) => Ok(res),
            Err(source) => Err(Report {
                source,
                source_code: Cow::Borrowed(&*self.ptr),
            }),
        }
    }
    // rest of methods which can mutate a pointer
}

pub struct Reporter<'p> {
    ptr: &'p Pointer,
}
// impl methods such as `assign`, `resolve`, etc

impl error::MutReporter<PointerBuf> for PointerBuf {
    fn report_err(&'_ mut self) -> Self::Reporter<'_> {
        MutReporter { ptr: self }
    }
}
```

asmello:
I don't follow what `Reporter` is playing here, but generally errors are intended to have static lifetimes, because they're expected to outlive their sources. This is why `Error::source` has `'static` in the return signature. So I think allocating-on-error is fine, actually. But I'm probably misunderstanding what you're trying to achieve.

chanced:
Yea, I should have explained that better. Sorry.

The reason `Reporter` is a trait is because I intend to implement it for `serde_json::Value` and `toml::Value`. All it does is provide a means to create a guard-like wrapper around a borrowed string or pointer that then re-exposes methods that it then converts into `Report`

```rust
pub struct Report<'a, E, T: ToOwned + ?Sized> {
    pub source: E,
    pub source_code: Cow<'a, T>,
}
```

re: allocation - are you saying all errors could/should allocate or simply that the above is unnecessary and all reports should owned?

asmello:
> do you mean all errors could/should allocate

Oh hell no, that would be an awful constraint. But they likely should have static lifetimes. So, for example, an error containing a static string is fine. In other cases the input/source can be owned. Allocate is a sort of last resort, but what I'm saying is that borrowing probably shouldn't even be attempted.

This doesn't preclude the usage of `Cow`, but I imagine we'd need `Cow<'static, T>` there ([not unheard of](https://github.com/metrics-rs/metrics/blob/d82bbb879f86ff5f5e9a21c5b3a3125f8154777e/metrics/src/common.rs#L18)). Unsure if it's worth it, but could be.

asmello:
As for the rest, thanks for the clarification. This sort of wrapping:

```rust
pub struct Report<'a, E, T: ToOwned + ?Sized> {
    pub source: E, // the error
    pub source_code: Cow<'a, T>, // str or Pointer
}
```

makes sense to me, although not sure how generalisable it is. Another thing to experiment with, I suppose.

chanced:
I'm game for dumping the `Cow` and making it an owned `String` or `PointerBuf`. I think it'll simplify borrowing constraints.

chanced:
@asmello i dont think this is going to work :(

I have a very little playground [here](https://github.com/chanced/spike-jsonptr-reports) where 2 flavors (`Mutable` and `Immutable`) are supposed to be in play but it doesn't seem to work how I want.

I was really hoping resolution wouldn't matter - that calling `report_err()` on multiple implementations would return which ever you ended up using (based on the methods used). That does not seem to be the case.

I'll toy around with it a bit more but I'm out of ideas if this doesn't end up working. I guess we just let them create the report by passing the error and `String`, `Pointer`, or `PointerBuf` (depending on context). 

```
git clone https://github.com/chanced/spike-jsonptr-reports
```

chanced:
I don't think this is worth it. I think we just provide a `Report` type that has a constructor that you pass an error and a PointerBuf and a string. The errors themselves contain the span.

chanced:
@asmello I refactored the errors, although I haven't gotten to the `Report` yet. I think it'll be pretty simple though.

This includes breaking changes that I have commented out for `ParseError`. Right now, I have `complete_offset`, `source_offset`, and `pointer_offset` commented out. I don't think we need them or the data needed to make them possible.

I can write up some top-level methods that we can provide to emulate them though. It'll just require the source string & the error.

I hate to break things like that though - I'm kinda banking on people having not had a need to use them, which is pretty weak.

chanced:
I should mention that I've got a lot cleaning up to do on it - I just let it sit idle for too long so I figured I'd push what I have. I plan on working on it over the weekend. 

The more I think about just dropping `offset`, the more I kinda regret that. I'll see if I can come up with a solution. I think I should also make the fields private.

If you'd rather keep them the way they are - as enum structs, I totally get that. This may be just too much.

chanced:
im going to start over, trying to minimize breaking changes.

PR #99, opened 2024-11-20
Experimental parse error from #93

chanced:
@asmello experimental parse error from #93  Sorry this took me so long to get back around to doing.

Its not as ergonomic as I had hoped, I forgot you can't provide defaults to generics of functions. Honestly, it may honestly be more hassle than its worth. I don't know if multi-error reporting is even remotely useful to folks for something as simple as a json pointer.

The miette output of the multi-error is a mess at the moment. I'm guessing its due to the related errors but I dont know how to get a list to show otherwise.

```rust
let err: ParseError<Complete> = PointerBuf::parse("hello-world/~3/##~~/~3/~") // or .parse_complete
    .unwrap_err() // ParseError<WithInput>
    .into(); // ParseError<Complete>

println!("{:?}", miette::Report::from(err));
```
Screenshot 2024-11-20 at 10 48 42 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Adds `Pointer::starts_with` and `Pointer::ends_with` for prefix and suffix matching.
 -   Adds new `ParseIndexError` variant to express the presence non-digit characters in the token.
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
--   Adds `ParseBufError`, returned as the `Err` side of `PointerBuf::parse`, which includes the input `String`.
+-   Adds `ParseBufError`, returned as the `Err` variant of `PointerBuf::parse`, which includes the input `String`.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Adds `Pointer::starts_with` and `Pointer::ends_with` for prefix and suffix matching.
 -   Adds new `ParseIndexError` variant to express the presence non-digit characters in the token.
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
--   Adds `ParseBufError`, returned as the `Err` variant of `PointerBuf::parse`, which includes the input `String` (from `Into::into`).
+-   Adds `InvalidEncoding` enum and corresponding field to `EncodingError` to distinguish between tilde and slash encoding errors
 
 ### Changed
 
@@ -24,6 +24,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   `Pointer::get` now accepts ranges and can produce `Pointer` segments as output (similar to
     `slice::get`).
 -   `PointerBuf::parse` now returns `BufParseError` instead of `ParseError`.
+
+### Deprecated
+
+-   `assign::AssignError` renamed to `assign::Error`.
+-   `resolve::AssignError` renamed to `resolve::Error`.
+-   `InvalidEncodingError` renamed to `EncodingError`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,17 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-
 ## [Unreleased]
 
 ### Added
 
--   Adds method `into_buf` for `Box<Pointer>` and `impl From<PathBuf> for Box<Pointer>`. 
+-   Adds method `into_buf` for `Box<Pointer>` and `impl From<PathBuf> for Box<Pointer>`.
 -   Adds unsafe associated methods `Pointer::new_unchecked` and `PointerBuf::new_unchecked` for
     external zero-cost construction.
 -   Adds `Pointer::starts_with` and `Pointer::ends_with` for prefix and suffix matching.
 -   Adds new `ParseIndexError` variant to express the presence non-digit characters in the token.
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
+-   Adds `ParseBufError`, returned as the `Err` side of `PointerBuf::parse`, which includes the input `String`.
 
 ### Changed
 
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Bumps minimum Rust version to 1.79.
 -   `Pointer::get` now accepts ranges and can produce `Pointer` segments as output (similar to
     `slice::get`).
+-   `PointerBuf::parse` now returns `BufParseError` instead of `ParseError`.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Adds `Pointer::starts_with` and `Pointer::ends_with` for prefix and suffix matching.
 -   Adds new `ParseIndexError` variant to express the presence non-digit characters in the token.
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
--   Adds `ParseBufError`, returned as the `Err` variant of `PointerBuf::parse`, which includes the input `String`.
+-   Adds `ParseBufError`, returned as the `Err` variant of `PointerBuf::parse`, which includes the input `String` (from `Into::into`).
 
 ### Changed
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,6 +18,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d4c819a1287eb618df47cc647173c5c4c66ba19d888a6e50d605672aed3140de"
 
 [[package]]
+name = "cfg-if"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
+
+[[package]]
 name = "env_logger"
 version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -39,7 +45,7 @@ version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ee8025cf36f917e6a52cce185b7c7177689b838b7ec138364e50cc2277a56cf4"
 dependencies = [
- "cfg-if",
+ "cfg-if 0.1.2",
  "libc",
  "wasi",
 ]
@@ -70,6 +76,7 @@ checksum = "d2f3e61cf687687b30c9e6ddf0fc36cf15f035e66d491e6da968fa49ffa9a378"
 name = "jsonptr"
 version = "0.6.3"
 dependencies = [
+ "miette",
  "quickcheck",
  "quickcheck_macros",
  "serde",
@@ -96,7 +103,7 @@ version = "0.4.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "14b6052be84e6b71ab17edffc2eeabf5c2c3ae1fdb464aae35ac50c67a44e1f7"
 dependencies = [
- "cfg-if",
+ "cfg-if 0.1.2",
 ]
 
 [[package]]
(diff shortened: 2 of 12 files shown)


chanced:
@asmello I'm sorry, github is back to marking messages I haven't seen as read. I'm not sure what that's about.

> Most of the noise in the output is due to that bug I flagged on https://github.com/chanced/jsonptr/pull/93, which we can just fix, and maybe we can tune the labels a bit for the multi-errors case, but it doesn't look terrible.

Yep, this was due to me delegating again. I created an enum `InvalidEncoding` and attached it to `EncodingError` (renamed from `InvalidEncodingError`). 

Single error display:
Screenshot 2024-11-21 at 11 02 00 AM

multi-error display (slightly different from above as I omitted the leading **_slash_**:
Screenshot 2024-11-21 at 11 02 37 AM

There's still a bug that trailing `~` are not displayed. The problem is the span length, i believe. 

> That said, I think we should leave it for a later step, after we've figured out the basic interface for rich errors.

Heh, I should have kept reading before investigating and writing code.

> Assuming we do this, my take is we shouldn't store the full offsets in the original error, and instead should compute them on-demand when the rich error is constructed (or even when .related_errors() is called!). It'd look like this from a usage perspective:

That random function attached to `Pointer` that didn't seem connected to the previous changeset was for this purpose. I started by computing them with the expectation that they would be either computed on the fly or dumped into `Subject` (because `offset` is relevant for `ParseError` but not `{assign, resolve}::Error`. I ended up just leaving them and it made things cleaner.

I'm not opposed to removing them from `assign::Error` and `resolve::Error`. When it comes to `ParseError`, I think it has gotten better now that the errors are being sourced properly. The verbiage of the errors themselves could use improvement and perhaps using `complete_offset()` instead of giving the offset of the token makes more sense.

> That usually goes for making types stateful, so that the internal data representation can be the same but you get a different set of methods available for each concrete marker choice.

There are a different set of methods based upon the associated types.

```rust
impl<S> ParseError<S> where S: Structure<Cause = Causes> {
    pub fn causes(&self) -> & [Cause]
}
impl<S> ParseError<S> where S: Structure<Cause = Cause> {
    pub fn cause(&self) -> &Cause
}
impl<S> ParseError<S> where S: Structure<Subject = String> {
    pub fn subject(&self) -> &str
}
```
`cause` and `subject` were `S::Cause` and `S::Subject` but they evaluated to the same (well, sort `&String` but close enough).

chanced:
I had really hoped this would be possible:
```rust
let Result<_, ParseError<Complete>> = Pointer::parse("/ptr");
```
where
```rust
impl Pointer {
    pub fn parse<S: AsRef<str> + ?Sized, E: Structure>(s: &S) -> Result<&Self, ParseError<E>>
}
```
but not being able to set defaults on generics of functions put an end to that. 

The advantage to using this structure is that the api of the error remains consistent between `Pointer::parse` and `PointerBuf::parse`. I haven't put any effort into going from `ParseError<WithoutInput>` -> `ParseError<WithInput>` but `ParseError<WithInput>` -> `ParseError<Complete>` is implemented (poorly). `ParseError<WithInput>` -> `ParseError<WithoutInput>` should also be implemented.

`Report` or whatever would still need to exist for `{assign, resolve}::Error` but we can feature flag it. Then we can `impl From<ParseError<Subject=String>> for Report` and the same machinery to go from `{assign, resolve}::Error` to `Report` will be possible for `ParseError<String>`

chanced:
Just to clarify on the design, there are 3 shapes that `ParseError` can be in (i'm not crazy about the names):
- `WithoutInput`, returned from `Pointer::parse`: `{ subject: Empty, cause: Cause }`
- `WithInput`, returned from `PointerBuf::parse`: `{ subject: String, cause: Cause }`
- `Complete`, bikeshed on where it is returned: `{ subject: String, cause: Causes }`

where `Cause` is formerly `ParseError` and `Causes` is `Box<[Cause]>`.
```rust
pub enum Cause {
    NoLeadingBackslash,
    InvalidEncoding {
        offset: usize,
        source: EncodingError,
    },
}
```

chanced:
If we had specialization, the API could be identical across variations of `ParseError`:

```rust
impl<S> ParseError<S> where S: Structure<Cause = Causes> {
    pub fn cause(&self) -> &Cause { &self.cause[0] }
}
```

chanced:
Nevermind, I was able to get `fn cause(&self) -> &Cause` to be universal without it.

```diff
pub trait Causative:
    PartialEq + Sized + std::error::Error + fmt::Display + fmt::Debug + miette::Diagnostic
+    fn first(&self) -> &Cause;
```

```rust
impl<S: Structure> ParseError<S> {
    pub fn cause(&self) -> &Cause {
        self.cause.first()
    }
}


```

chanced:
An alternative approach would be something like
```rust
// using struct variants here to keep the example simple.

struct SansInput { 
    cause: Cause, 
}
struct WithInput { 
    cause: Cause, 
    input: String
}
enum ParseError {
    SansInput(SansInput),
    WithInput(WithInput),
    Complete(Complete),
}
```
but then methods like accessors for the input would need to return `Option`.

asmello:
I'm still skeptical about this marker trait system, because in my experience this pattern has some serious ergonomics implications. One you kinda noticed already - lack of specialisation means we often have to implement the same function multiple times, once for each concrete marker value. This has a viral effect, because each function that is implemented concretely can't be called in generic contexts. That is, this doesn't work:

```rust
use std::marker::PhantomData;

trait State {}
struct S1;
impl State for S1 {}
struct S2;
impl State for S2 {}
struct Foo<S: State> {
    _phantom: PhantomData<S>
}

impl Foo<S1> {
    fn do_it() { }
}

impl Foo<S2> {
    fn do_it() { }
}

fn frobnicate<S: State>(foo: Foo<S>) {
    // error[E0599]: no method named `do_it` found for struct `Foo<S>`
    foo.do_it(); 
}
```

[Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=3fc3b973654a2dfec2871945377cf6df)

I'm not actually sure how your `first` method in `Causative` solved this issue? But even if it did, you still went from implementing `cause` 3x to implementing `first` 3x, which doesn't seem like much of an improvement. What if you needed `last` as well? Doesn't seem to scale well.

I'm also worried that all this work is just to support `ParseError`, and I'm not sure how much can be reused for other error types that would similarly benefit from an enrichment system.

asmello:
To be clear though, you are not crazy, there is some credit to this system. The state machine is a nice model for error enrichment. My concerns are just over implications for ergonomics, as well as complexity.

I'm also scared of running into edge cases with generics and associated types. Fought those hard last Sunday getting `.into_owned` to work when I had a very similar design to `ParseError` (where a field's type was an associated type). The solution ended up being to be more loose with the generic bounds on the type definition (introducing a totally decoupled second parameter) and then constraining it just in the places where the association was relevant. That led to many simplifications. I wonder if something similar might be done here.

asmello:
> The remaining advantage to using this structure is that the api of the error remains consistent between Pointer::parse and PointerBuf::parse.

I wonder if the answer isn't to just have 3 totally separate types, but have them implement a trait where common functionality makes sense? This would give callers a common API to operate with where it matters, while still providing a way to have different methods on each stage of enrichment.

What's telling to me here is that the three flavours of `ParseError` end up having _both_ distinct data types as well as a unique subset of methods. I really don't see what the generics are buying us here.

asmello:
> Understood and thanks. I wasn't sure if you outright hated it or not :).

Oh, I don't hate it. I'm just scared of it! 😄 

> Ah, I hadn't really considered too much the implications of infecting generics on people. Hmm.

I _think_ this is probably ok for users, because they have to specify which concrete type they want out of `.parse`, and once you've got a concrete type it's all good. It's only if they decide to be generic themselves over `S` that they will start to run into trouble, but I think this would be rare. I'm mainly concerned about the cost of internal evolution.

chanced:
> I wonder if the answer isn't to just have 3 totally separate types, but have them implement a trait where common functionality makes sense? This would give callers a common API to operate with where it matters, while still providing a way to have different methods on each stage of enrichment.

That may work. If we go that route, I'm thinking we eliminate the third option so it'll just be 2, with and without input. 

> What's telling to me here is that the three flavours of ParseError end up having both distinct data types as well as a unique subset of methods. I really don't see what the generics are buying us here.

Yea, we have 2 or 3 distinct structures depending on whether we keep the list. The generics keep it to one descriptive type. That's about it. Unfortunately the luster of this approach is severely diminished without being able to pick and choose the reporting style from a single set of methods (`parse` on both `Pointer` and `PointerBuf`) by specifying the generic.

> Oh, I don't hate it. I'm just scared of it! 😄

Hah, I understand that. `grill` has become a swamp of associated types and generics (largely thanks to `slotmap::SlotMap<K: Key>`). Its the only other project I have which is infested with associated types. I've come to love them but booting back up on that project takes time.

Having said all of that, I don't think this particular arrangement is horrible, especially if we seal up `Causitive` (with a better name) but its still mental overhead. Also, there are so few variants that it seems a bit overkill.

chanced:
I think I'd favor a distinct type for the `ParseError` with the input string. `Report<ParseError>` may be too generic for an explicit action.

chanced:
I'll be awol this weekend but I think I'm leaning toward the wrapper as well. The alternative is going to be basically the equivalent of it anyway.

Basically the two options are:
```rust
enum ParseError {}
struct EnrichedParseError { input: String, error: ParseError }
```
or something like this to get the same API across types:
```rust
enum Cause {}
struct ParseError { cause: Cause }
struct EnrichedParseError { cause: Cause, input: String }
```

Both of which don't make a whole lot of sense if we have a wrapper type.

asmello:
Taking a step back to re-evaluate the overall approach with a fresher perspective.

For rich errors (including multi-errors) in the parsing API, I think the right approach would be to introduce separate parsers for each level of error richness. This way we keep the core API simple and avoid doing extra work without giving up the option to produce the enhanced errors when it matters.

```rust
impl Pointer {
    pub fn parse(s: &str) -> Result<Self, ParseError> {...}
}

impl PointerBuf {
    pub fn parse(s: impl Into<Cow<'_, str>>) -> Result<Self, ParseError> {...}
}

trait Parser {
    type Error;
    pub fn parse(s: impl Into<Cow<'_, str>>) -> Result<Cow<'_, Pointer>, Self::Error>;
}

let err: ParseError = SimpleParser::parse("foo").unwrap_err();
let err: RichParseError = RichParser::parse("foo").unwrap_err();
let err: FullParseError = FullParser::parse("foo").unwrap_err();
```

I'm keeping the `Cow` signature in `PointerBuf` for the ergonomics. It'd be more explicit to require `Pointer::parse("/foo").to_buf()` as the way to go from reference to owned, but alas it's annoying to add the `.to_buf()` which is why we moved to `Into<String>` in the first place.

For the `Assign` error, it doesn't really make sense to have multi-errors, as far as I can see. So the helper struct makes a bit less sense. That said, it also gives us an opportunity to also allow users to configure how they want to deal with missing nodes along the path. I know you've explored this in the past, but unless my memory fails we haven't tried this as a separate struct yet, right? I'm thinking of HTTP clients here — where you configure your request parameters, then execute the request —, except we configure the assignment strategy and then execute it. The details there are for a separate discussion, but wanted to bring up the side benefit.

For resolve, it makes even less sense to have a helper struct, since there's nothing really to configure, but may still be worth it for the rich error interface. Multi-errors also don't make sense here.

Another option for assign and resolve is to just condition the error types on the `miette` flag, and have it always be rich when miette integration is enabled. There isn't much of a runtime cost, really, it's just a difference in API. Doing it like this is more annoying for us from a code organisation and testing perspective, but it's the most straightforward for users.

Multi-errors could make sense in `Index` parsing, but not sure it's worth it. Otherwise it's the same as the resolve errors.

You mentioned being away this weekend, so no pressure to reply. I'll probably come back to all of this next week(end?) after I've given it all some more thought.

asmello:
Also worth mentioning, the one issue with the wrapper type, even if we make it work with multi-errors, is the potential of mismatch between the subject and the error. It's even weirder with multi-errors because, to generate them, we'd need to essentially run another full scan and may find entirely different errors if given a different source. This is what made me feel like it's the wrong direction to take. Ideally we don't need to revisit the subject other than for display purposes, once an error has been generated.

I said before the mismatch is a low risk given typically you'd use the wrapper immediately like `Pointer::parse(s).diagnose(s)`, but it smells, and I don't like it. The `Parser` approach seems safer without compromising ergonomics, and I think it's a bit easier to evolve, though it requires a bit more boilerplate. That said, I may still change my mind if other factors surface.

chanced:
Ohhh!! Now that's a clever way to handle it!

Sorry, catching up now.

chanced:
> For rich errors (including multi-errors) in the parsing API, I think the right approach would be to introduce separate parsers for each level of error richness. This way we keep the core API simple and avoid doing extra work without giving up the option to produce the enhanced errors when it matters.

Ohh! Now that's a clever way to handle it! I really like the approach of having different traits. It solves what this design failed to do. That is, allow for structuring the error by specifying a call without cluttering up the API with numerous methods.

>
```rust
impl PointerBuf {
    pub fn parse(s: impl Into<Cow<'_, str>>) -> Result<Self, ParseError> {...}
}
```
What if we always allocate for rich and full parser? It seems like if you want an error with the input, you are on an error path and allocating a string is fine, like you were saying earlier.

Digesting the rest.

chanced:
> For the Assign error, it doesn't really make sense to have multi-errors, as far as I can see. So a helper struct like Assigner makes a bit less sense. 

Right, `{assign, resolve}::Error` really don't need a wrapper error. The wrapper class did provide a means to report the specific error and underline the token.

> That said, it also gives us an opportunity to also allow users to configure how they want to deal with missing nodes along the path. I know you've explored this in the past, but unless my memory fails we haven't tried this as a separate struct yet, right? I'm thinking of HTTP clients here — where you configure your request parameters, then execute the request —, except we configure the assignment strategy and then execute it. The details there are for a separate discussion, but wanted to bring up the side benefit.

That sounds like an awesome idea. I'm game for trying it in the future. I would love something like it.

chanced:
superseded by #101

PR #100, opened 2024-11-28
refactor a bit

chanced:
Hey @asmello, I made some changes for you to review. High level:

- renames `SRC` generic of `Report`  to `D` and bound it to `Diagnostic`
- drops `SUB` in generic, uses `D::Subject` instead
- removes lifetime from `Report`
- removes seal for `Diagnostic`
- ~~Adds associated type `Related` and method `related` to `Diagnostic` for multi-errors~~

Regarding lifetimes, I really think we should only deal with owned subjects. It simplifies the code considerably, reduces boilerplate, avoids `borrowck` issues, and doesn't cost much; a single allocation at worst on an error path for reporting seems more than fair. Given that this behavior is opt-in, I say we keep it simple.

I really don't know about the multi-error list. We can include it and I made some changes to support it, but I don't know how valuable it'll be. However, it is possible with this structure. To support it, we will need a `struct ParseErrors(Box<[ParseError]>` that would have a `Related = ParseError` and I'll need to move over the logic from the other branch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Adds new `ParseIndexError` variant to express the presence non-digit characters in the token.
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
 -   Adds `ParseBufError`, returned as the `Err` variant of `PointerBuf::parse`, which includes the input `String` (from `Into::into`).
+-   Adds `InvalidEncoding`
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 -   Adds `Pointer::starts_with` and `Pointer::ends_with` for prefix and suffix matching.
 -   Adds new `ParseIndexError` variant to express the presence non-digit characters in the token.
 -   Adds `Token::is_next` for checking if a token represents the `-` character.
--   Adds `ParseBufError`, returned as the `Err` variant of `PointerBuf::parse`, which includes the input `String` (from `Into::into`).
--   Adds `InvalidEncoding`
+-   Adds `InvalidEncoding` to represent the two possible encoding errors when decoding a token.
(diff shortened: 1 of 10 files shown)


chanced:
Sorry I went awol on this. I'll be heading home tomorrow and this will be my first priority.

chanced:
@asmello So the time I had allocated after my trip to work on this was unfortunately spent sick. I have had time to tinker with it here and there but I'm struggling to nail down a design with regard to mutli-errors.

The biggest hurdle is that [`miette::Diagnostic::related`](https://docs.rs/miette/latest/miette/trait.Diagnostic.html#method.related) expects an return of an optional `Iterator<Item = &'a dyn miette::Diagnostic>`.

I considered `OnceLock` + a borrowed flavor of `Report` but that doesn't work as they cannot be generated on the fly and returned. The only way to properly display them, so far as I can reason, is cloning the subject/sourcecode for each error - albeit lazily (via `OnceLock`). That way the related errors can satisfy `miette::Diagnostic` and display the pointer (related errors are displayed as an independent error, disconnected from the original).

I'm at the point where I'm pretty certain that related errors just aren't worth it. It spikes the complexity considerably with little upside (if a pointer is malformed, its malformed and as pointers are likely something being generated by logic, i don't know how many users would really benefit from having a detailed demonstration.

I think everything else is solid and I'll finish it out, hopefully sometime this weekend. We can decide whether to proceed without them from there.

chanced:
There's an issue with reporting being displayed twice, need to fix that:
Screenshot 2025-01-27 at 12 27 12 PM

chanced:
I "fixed" the issue of repeated sourcing by returning `self.source.source()` from `Error::source` on `Report`. I experimented with a whole bunch of different options but I cant think of any other way to go about it. I thought I had something with [`miette::Diagnostic::diagnostic_source(https://docs.rs/miette/latest/miette/trait.Diagnostic.html#method.diagnostic_source) but that return `-> Option<&dyn Diagnostic>` really put a kink in creating lifetime toting enums for specifically for that purpose.

I modified the error messages such that each error does not delve into the source, displaying only information at that depth:

Screenshot 2025-01-29 at 11 22 37 PM
Screenshot 2025-01-29 at 11 22 18 PM
Screenshot 2025-01-29 at 11 34 31 PM

PR #103, opened 2025-02-16
make `PointerBuf::parse` error basic by default

asmello:
Since #101 the `assign` and `resolve` `Error` types are self-contained and
shallow by default (meaning they don't include a copy of their source string).
This is nice - the errors are helpful and don't require extra allocations, but
the users may still opt into richer errors by using the `Diagnose` trait and
providing the "source code" (in `miette` terminology).

The `PointerBuf::parse` error was different, however. It was made a rich error
by default, which requires one extra clone (which _may_ be unnecessary), but
more importantly, is inconsistent with the other error contract.

This PR makes the parse error consistent, by switching it back to `ParseError`.
To recover the rich error functionality users can just call `.diagnose()` on the
`Result`, same as with `assign::Error` and `resolve::Error`.

I'm also sneaking in a fix for the eager behaviour of `Diagnose::diagnose_with`.


asmello:
Aha, I think I figured out (rediscovered?) why `PointerBuf::parse` was special cased. It means when using the rich error interface, then allocation always take place. This is not too unreasonable, but I'll put some thought into whether it can be avoided.

asmello:
A couple options:
1. Make the error an enum:
```rust
enum ParseErrorWrapper {
    Simple(ParseError),
    Rich(Report<ParseError)
}
```
and pivot based on the `Cow` variant. With this it makes sense to actually have a `Cow` rather than use the `CowStr` trait.
2. Optionally keep the `String` inside `ParseError`, then make it so `into_report` can use the internal subject rather than a provided one. This would require a trait refactor, not sure how it'd look exactly. It also means making `ParseError` a struct so we can hide the `String` field (not strictly necessary, but I think it's a good idea).

Either way there's some amount of user-facing complexity added, but on the other end this is kinda dumb:

```rust
    let ptr_str = "hello";
    PointerBuf::parse(ptr_str).diagnose_with(|| ptr_str.to_owned()); // kinda fine
    let ptr_str = String::from(ptr_str);
    PointerBuf::parse(&ptr_str).diagnose(ptr_str); // dumb, we allocate in Ok case unecessarily
```

Ugh, I may reconsider this PR entirely.

asmello:
Ok, let's drop this. I'll make a different PR with the side-benefits of this one.

asmello:
Side-benefits over at https://github.com/chanced/jsonptr/pull/104

PR #104, opened 2025-02-17
misc fixes and nits from clippy

asmello:
- Make `diagnose_with` lazy, as it was intended to be
- Clippy nits related to lifetimes
- remove leftover test


asmello:
Going to merge since I think this is fairly uncontroversial, but worth a look for awareness @chanced .

PR #105, opened 2025-02-17
clean up misc from retro-review of #101

asmello:
I changed my mind on removing serde as a default feature. It _is_ a heavy dependency to bring by default, but it's easy enough to disable the default features, and I do see the "common" use case for jsonptr involving `serde_json`. Also I guess `Diagnose` makes sense when you consider that the error it converts isn't _yet_ a `Report`, even though it implements `Diagnostic`. Perhaps it's `Diagnostic` that requires reconsidering.

Changes I'm pushing forward:
- make `diagnostic::Label::new` private
- add docs do `validate_bytes`
- remove `NoLeadingSlash` struct
- update `delete.rs` docs
- deprecate `ResolveError`

Left for standalone PRs:
- Migrate `index::ParseIndexError` and `token::EncodingError` to the new error system.
- Seal `Diagnose`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+-   Made `Label` type crate-private.
+
+### Deprecated
+
+-   Deprecated `ResolveError` name.
+
 ## [0.7.1] 2025-02-16
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
--   Made `Label` type crate-private.
-
 ### Deprecated
 
 -   Deprecated `ResolveError` name.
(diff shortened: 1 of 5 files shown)


asmello:
This is technically breaking, as reported by the semver check, but nobody would be using the removed type because it's useless / not referenced anywhere.

chanced:
> make diagnostic::Label::new private
This one needs to be public in order for users to implement `Diagnostic` properly. Users can bring their own errors for `Assign` and `Resolve` but they have to implement `Diagnostic`.

> remove NoLeadingSlash struct
Ugh, thank you. Sorry I left that in.

asmello:
> This one needs to be public in order for users to implement Diagnostic properly. Users can bring their own errors for Assign and Resolve but they have to implement Diagnostic.

Ah, interesting, but do we expect users to implement `Diagnostic` themselves?

chanced:
I dont necessarily expect it - I doubt anyone is implementing `Assign` and `Resolve` for custom languages or data structures - but it is possible that someone to do so and in an effort to facilitate that, the traits were designed so that implementations bring their own error types for `resolve::Resolve` and `assign::Assign`.

asmello:
You know what, you're completely right. I lost track of what those traits were for, it does make sense that people can use `jsonptr` with their own types. Although, now that I think about it, it also only makes sense for JSON-like data structures.

I'll update this PR and remove that change. Thanks!

asmello:
Since you haven't explicitly nacked, and other than the deprecation this is mostly docs changes, I'm going to proceed with merging. 👍

PR #107, opened 2025-02-17
seal the `Diagnose` trait

asmello:
To prevent users from implementing it for their types (which doesn't really make sense). This protects us from future breaks.


asmello:
Semver rightfully reports this as a break, but better do it now than post-1.0.

PR #108, opened 2025-02-18
migrate remaining errors to new model

asmello:
The `EncodingError` and `ParseIndexError` types are parsing errors with a
well defined 'subject', but they weren't previously made into `Diagnostic`
implementors, so they worked a bit inconsistently with other error types. This
PR makes it so those errors follow the same conventions and general structure
used by its kindred.

There is one related breaking change: `InvalidCharacterError` previously held
a copy of the subject, but in the new model this is a responsibility of its
enriched `Report<InvalidCharacterError>` type. Since some methods directly
depended on access to the subject, they had to be removed. I think this is
justifiable since the new error APIs are already breaking (we just should've
made this change in tandem with the other related breaks).

I don't have any other error-related breaks planned.

This PR also incidentally fixes an incorrect error type returned if the token ended with a `~` (it would be flagged as a slash error, not tilde). I updated tests to cover this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,23 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


chanced:
Damnit, I really regret pushing `0.7`. Agreed that it should have been bundled.

We can cut a `0.8` in short order i think. I doubt many will have picked up `0.7` already and the breaks slotted for it won't be that bad.

PR #110, opened 2025-02-23
improve `Token::from_encoded`

asmello:
Non-breaking change. Accept owned or borrowed strings as input in `Token::from_encoded`.

### DIFF
--- a/src/token.rs
+++ b/src/token.rs
@@ -70,9 +70,10 @@ impl<'a> Token<'a> {
     /// ## Errors
     /// Returns `InvalidEncodingError` if the input string is not a valid RFC
     /// 6901 (`~` must be followed by `0` or `1`)
-    pub fn from_encoded(s: &'a str) -> Result<Self, EncodingError> {
+    pub fn from_encoded(s: impl Into<Cow<'a, str>>) -> Result<Self, EncodingError> {
+        let inner = s.into();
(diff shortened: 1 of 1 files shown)


chanced:
This is breaking but probably pretty uncommon. A type could implement `Deref<Target = str>` but not `Into<Cow<str>>`

```rust
struct X;

impl std::ops::Deref for X {
    type Target = str;
    fn deref(&self) -> &Self::Target { "" }
}

fn ok(v: &str) {}
fn breaking<'a>(v: impl Into<std::borrow::Cow<'a, str>>) {}

fn main() {
    ok(&X);
    breaking(&X);
}
```
[playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=38ffa9b773476a646305b6e92ee7c3ae)

If memory serves, we have breaking changes in main already so I'm pulling it in.

asmello:
Huh, fascinating! I honestly didn't know Deref worked like that! I thought the only type that satisfied a `&str` signature was, well, `&str`.

In that case, let's document this as breaking accordingly and make sure to include in the next minor release. Thanks for flagging.

asmello:
Actually, I think the changelog is fine as is. Since we have harder unreleased breaks, I'll assume you're ok with this and will proceed with merging.

PR #112, opened 2025-02-26
README: Fix `Delete` example

brooswajne:
Seems like this was accidentally left as a copy-paste of the `Assign` example. I took the liberty to add some extra keys to the object which we're deleting from, as I think it makes it a bit clearer what's happening than:

    { "secret": { "universe": 42 }} -> { "secret": {} }

Which confused me a bit at first (but that might just be me).

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,10 +136,10 @@ use jsonptr::Pointer;
 use serde_json::json;
 
 let ptr = Pointer::parse("/secret/universe").unwrap();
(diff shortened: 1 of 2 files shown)


asmello (approved):
Perfect! Thank you very much for noticing and contributing.

asmello:
I guess semver will continue failing until we make a new release. @chanced 

Since this is fairly uncontroversial I'll go ahead and merge.

